PR #2, opened 2025-11-13
Add BibleWidgetView

bmanquen:
### Add BibleWidgetView component and enhance usePassage hook with heading/notes options

  Summary

-  Add new BibleWidgetView component for displaying Bible passages in a widget format
-  Enhance usePassage hook to support include_headings and include_notes options
-  Add Storybook story for the new component with light and dark mode examples

### DIFF
--- a/packages/ui/src/components/bible-widget-view.tsx
+++ b/packages/ui/src/components/bible-widget-view.tsx
@@ -39,8 +39,8 @@ export function BibleWidgetView({
         versionId={versionId}
       />
 
-      <div className="yv:grid yv:grid-cols-2 yv:items-center yv:mt-8">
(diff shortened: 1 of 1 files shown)


packages/ui/src/styles/bible-reader.css
```diff
@@ -24,10 +24,22 @@
...
-  
+
   /* We want to make sure that the paragraph is not too wide */
   @apply yv:*:w-full yv:text-primary yv:*:max-w-lg;
+}
+
+/* Dark theme override for Bible text */
+[data-yv-theme="dark"] [data-slot="yv-bible-renderer"] {
+  @apply yv:text-foreground;
+}
```
cameronapak: issue: This may not be needed. I think all we need to do to update verse.tsx to use `yv:text-primary-foreground` and it should work for light and dark mode.

PR #4, opened 2025-11-13
Set up automated NPM publishing

hagmic:
Implements a complete CI/CD pipeline for the YouVersion Platform SDK monorepo with automated NPM publishing on merge to main.


hagmic:
> Hey @hagmic, thanks for working on the publishing!
> 
> In our migration yesterday, I forgot to move over our GitHub workflows and the prettier config. Can you look at #3 ? I mention this because some of the code will conflict

Hey @cameronapak! David asked me to configure NPM publishing, but didn't realize you guys had already done some setup. Comparing the two branches, it looks like the only piece this PR is missing the storybook/chromatic jobs.

This PR is using some newer versions of the actions, what do you think of me adding those missing jobs to this PR and removing them from Brendan's? Open to whatever is easiest.

hagmic:
@jhampton I requested access to that doc, didn’t realize it existed!

Happy to make any changes requested.

jhampton:
> @jhampton I requested access to that doc, didn’t realize it existed!
> 
> Happy to make any changes requested.

Thank you!  I just added you to the space (sorry about that!).  Aside from enforcing conventional commits, your approach is already aligned.

jhampton:
@hagmic I just spoke with @davidfedor about our `versioning` strategy.  There are two paths:

1.  Each package is versioned independently, and the packages _must_ depend on the latest version of other packages, especially in the case of `major` version bumps, explicitly

or

2. All packages are versioned to the same, _highest_ version and automatically depend on the latest version of other packages.

The tradeoffs are somewhat equivalent, and this choice _mostly_ affects how we think about communicating changes and what-needs-updating-when moving forward.  We'd love to get your opinion since you all have more experience and may have future maintainer responsibility.  Once we know the path forward, we can document it, socialize it, and move on.

---

Claude's comment (which seems sensible) about tradeoffs:

### ✅ Advantages of Using Single Version

1. **Simplicity:** One version number to track
2. **Consistency:** All packages always in sync
3. **Predictability:** Users know all packages work together
4. **Clear releases:** Each release is a cohesive unit
5. **Easier dependency management:** No version mismatch issues

### ⚠️ Trade-offs

1. **More frequent releases:** Even small changes trigger version bump for all packages
2. **Larger version numbers:** Versions increment faster than with independent versioning
3. **Unchanged packages publish:** Packages republish even without changes (npm handles this well)

---

_Alternatively_, we could decide (**in the future**) to publish a _single_ package per-repo and ensure that consumers can tree-shake unused portions as the SDK grows in size.  Based on a cursory review of our bundling, consumers will end up with all imported code and will have **Limited Tree-Shaking** capability:

### Key Issues:

1. **`splitting: false` in tsup config** (`packages/ui/tsup.config.ts:7`)
   - Your UI package bundles everything into single `index.js` and `index.cjs` files
   - Without code splitting, bundlers like webpack/vite cannot eliminate unused code paths effectively

2. **`noExternal: ['@youversion/platform-core']`** (`packages/ui/tsup.config.ts:11`)
   - Core package is bundled directly into the UI package
   - Consumers get the entire core library even if they only use UI components

3. **Barrel export pattern** (`packages/ui/src/index.ts`)
   - Single entry point with `export *` re-exports everything
   - While ESM supports tree-shaking with barrel exports, the bundled output negates this benefit

4. **Bundled dependencies create a monolithic artifact**
   - A consumer importing just `SignInButton` still downloads code for `BibleChapterPicker`, `VerseOfTheDay`, all core API clients, etc.

hagmic:
> @hagmic I just spoke with @davidfedor about our `versioning` strategy. There are two paths:
> 
> 1. Each package is versioned independently, and the packages _must_ depend on the latest version of other packages, especially in the case of `major` version bumps, explicitly
> 
> or
> 
> 2. All packages are versioned to the same, _highest_ version and automatically depend on the latest version of other packages.
> 
> The tradeoffs are somewhat equivalent, and this choice _mostly_ affects how we think about communicating changes and what-needs-updating-when moving forward. We'd love to get your opinion since you all have more experience and may have future maintainer responsibility. Once we know the path forward, we can document it, socialize it, and move on.
> 
> Claude's comment (which seems sensible) about tradeoffs:
> 
> ### ✅ Advantages of Using Single Version
> 1. **Simplicity:** One version number to track
> 2. **Consistency:** All packages always in sync
> 3. **Predictability:** Users know all packages work together
> 4. **Clear releases:** Each release is a cohesive unit
> 5. **Easier dependency management:** No version mismatch issues
> 
> ### ⚠️ Trade-offs
> 1. **More frequent releases:** Even small changes trigger version bump for all packages
> 2. **Larger version numbers:** Versions increment faster than with independent versioning
> 3. **Unchanged packages publish:** Packages republish even without changes (npm handles this well)
> 
> _Alternatively_, we could decide (**in the future**) to publish a _single_ package per-repo and ensure that consumers can tree-shake unused portions as the SDK grows in size. Based on a cursory review of our bundling, consumers will end up with all imported code and will have **Limited Tree-Shaking** capability:
> 
> ### Key Issues:
> 1. **`splitting: false` in tsup config** (`packages/ui/tsup.config.ts:7`)
>    
>    * Your UI package bundles everything into single `index.js` and `index.cjs` files
>    * Without code splitting, bundlers like webpack/vite cannot eliminate unused code paths effectively
> 2. **`noExternal: ['@youversion/platform-core']`** (`packages/ui/tsup.config.ts:11`)
>    
>    * Core package is bundled directly into the UI package
>    * Consumers get the entire core library even if they only use UI components
> 3. **Barrel export pattern** (`packages/ui/src/index.ts`)
>    
>    * Single entry point with `export *` re-exports everything
>    * While ESM supports tree-shaking with barrel exports, the bundled output negates this benefit
> 4. **Bundled dependencies create a monolithic artifact**
>    
>    * A consumer importing just `SignInButton` still downloads code for `BibleChapterPicker`, `VerseOfTheDay`, all core API clients, etc.

@jhampton  For now I would recommend unified versioning with independent packages: Unified versioning makes sense mostly because developers don't want to be figuring out "which version of UI works with which version of core?". It's also easier to communicate "YouVersion Platform SDK v2.0.0 was released with....". A downside to this is that we may end up publishing unchanged packages as new versions, but not  a big deal.

Keeping multiple packages makes sense because some users may only need/want the core library (Node or server use cases) and also gives us a clearer separation of concerns (data layer vs UI layer). I've updated the changesets config and documentation to reflect that decision.

cameronapak (approved):
LGTM.

PR #8, opened 2025-11-14
YPE-634 - BibleReader component

cameronapak:
> [!important]
> Watch the quick overview video here: https://share.cleanshot.com/t8DcnGxg

CleanShot 2025-11-14 at 16 08 44@2x

CleanShot 2025-11-14 at 16 08 47@2x

### DIFF
--- a/packages/ui/src/components/bible-version-picker.tsx
+++ b/packages/ui/src/components/bible-version-picker.tsx
@@ -115,13 +115,16 @@ function Root({
   );
 }
 
-export type TriggerProps = Omit<React.ComponentProps<typeof PopoverTrigger>, 'children'> & {
+export type BibleVersionPickerTriggerProps = Omit<
+  React.ComponentProps<typeof PopoverTrigger>,
+  'children'
+> & {
   children?:
     | React.ReactNode
     | ((props: { version: BibleVersion | null; loading: boolean }) => React.ReactNode);
 };
 
-function Trigger({ asChild = true, children, ...props }: TriggerProps) {
+function Trigger({ asChild = true, children, ...props }: BibleVersionPickerTriggerProps) {
   const { versionId, background } = useBibleVersionPickerContext();
   const { version, loading } = useVersion(versionId);
 
--- a/packages/ui/src/components/index.ts
+++ b/packages/ui/src/components/index.ts
@@ -1,6 +1,10 @@
 export { BibleChapterPicker, type RootProps, type TriggerProps } from './bible-chapter-picker';
 export { BibleReader, type BibleReaderRootProps } from './bible-reader';
-export { BibleVersionPicker, type BibleVersionPickerRootProps } from './bible-version-picker';
+export {
+  BibleVersionPicker,
+  type BibleVersionPickerRootProps,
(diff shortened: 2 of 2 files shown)


packages/ui/src/components/bible-reader.tsx
```diff
@@ -0,0 +1,214 @@
...
+import { Button } from './ui/button';
+import { cn } from '@/lib/utils';
+
+type BibleReaderContextType = {
+  book: string;
+  chapter: string;
+  versionId: number;
+  setBook: (book: string) => void;
+  setChapter: (chapter: string) => void;
+  setVersionId: (versionId: number) => void;
```
bmanquen: Should we match how React types setStates? Like this, `React.Dispatch<React.SetStateAction<string>>`

packages/ui/src/components/bible-reader.tsx
```diff
@@ -0,0 +1,214 @@
...
+          <p className="yv:text-balance yv:text-[0.75em] yv:text-center yv:text-muted-foreground">
+            {version.copyright_short}
+          </p>
+        </footer>
+      )}
+    </main>
+  );
+}
+
+function Toolbar({ border = 'top' }: { border?: 'top' | 'bottom' }) {
```
bmanquen: nit: do we need to pass the props as an object here since there is just one prop?

packages/ui/src/components/bible-version-picker.tsx
```diff
@@ -324,3 +333,4 @@ function Content() {
 }
 
 export const BibleVersionPicker = Object.assign({}, { Root, Trigger, Content });
+export type BibleVersionPickerRootProps = RootProps;
```
bmanquen: thought: do we need to export all prop types?

bmanquen (changes requested):
Great work!

packages/ui/src/components/bible-reader.tsx
```diff
@@ -0,0 +1,214 @@
+'use client';
+
+import { createContext, useContext, useMemo, type ReactNode } from 'react';
+import { useControllableState } from '@radix-ui/react-use-controllable-state';
```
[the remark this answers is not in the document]
cameronapak: I gently push back, because I do not think this is blocking. 

It says not intended, but 20 million weekly downloads of this npm package tells me it's safe. 

I use it because of https://www.components.build/state#merging-states

packages/ui/src/components/bible-reader.tsx
```diff
@@ -0,0 +1,214 @@
...
+          </BibleVersionPicker.Trigger>
+          <BibleVersionPicker.Content />
+        </BibleVersionPicker.Root>
+      </div>
+    </section>
+  );
+}
+
+export const BibleReader = Object.assign({}, { Root, Content, Toolbar });
+export type BibleReaderRootProps = RootProps;
```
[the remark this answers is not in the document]
cameronapak: I think so... https://github.com/youversion/platform-sdk-react/blob/main/packages/ui/src/components/index.ts seems to lean in that direction.

bmanquen (approved):

PR #24, opened 2025-11-17
chore(core): update language tests to use mws

birdwell:
Remove a dup test and a couple that are testing integration API things. 

Add mws support for the remaining tests.


cameronapak (approved):
LGTM. 

note: I recognize many vitest tests are failing. I assume this is due to the API being down

PR #30, opened 2025-11-18
test(core): mock all api endpoints used in tests

bmanquen:


packages/core/src/__tests__/authentication.test.ts
```diff
@@ -38,7 +38,7 @@ describe('AuthClient', () => {
         headers: {
           get: vi.fn((name: string) => (name === 'content-type' ? 'application/json' : null)),
         },
-      } as Response);
+      } as unknown as Response);
```
cameronapak: **question (non-blocking):** Did this work before as `as Response` versus `as unknown as Response`

packages/core/src/__tests__/handlers.ts
```diff
@@ -90,4 +104,88 @@ export const handlers = [
...
   }),
+
+  // Versions endpoints
+  http.get(`https://${apiHost}/v1/bibles`, () => {
+    return HttpResponse.json({
+      data: mockVersions,
+      next_page_token: null,
+      total_size: 14,
+    });
+  }),
```
cameronapak: **praise:** Nice. Good work on having the mock data separate to see the mocked endpoints so clearly

packages/core/src/__tests__/authentication.test.ts
bmanquen: It was but Typescript was giving me an error on it. I wonder if our Typescript doesn't check test files. Not sure why the typescript checks passed it.

PR #46, opened 2025-11-25
fix(ui): remove the need to import our css styles

bmanquen:
- Users had to import our css file and this caused post processing by the users build tools.
- This removes the need to import the css file and will fix the post processing issues that users were having.

This is a simplified solution based on the work that @cameronapak in #45.


mic-mart (commented):
See the codex comment on the PR and my response to it.
We need to ensure production builds won't break.

PR #68, opened 2025-12-14
feat(ui): Add footnotes to BibleTextView

bmanquen:
Single footnote:
image

Multiple footnotes:
image


mic-mart (approved):
Looks good for now. I think we have an opportunity to optimize the algorithm and protect against possible memory bloat with better cleanup, but let's move forward now and come back later if we need to optimize.

PR #86, opened 2025-12-20
feat(core): add pagination to getVersions endpoint

bmanquen:


packages/core/src/__tests__/languages.test.ts
```diff
@@ -32,7 +32,11 @@ describe('LanguagesClient', () => {
       const { success } = LanguageSchema.safeParse(languages.data[0]);
       expect(success).toBe(true);
       expect(languages.data).toHaveLength(20);
-      expect(languages.data.every((language) => language.countries?.includes('US'))).toBe(true);
+      /** This is not true, it will always return the number provided in page_size
+       *  and it will put the ones that have the country in their countries array
+       *  at the top of the list.
+       */
+      // expect(languages.data.every((language) => language.countries?.includes('US'))).toBe(true);
```
cameronapak: question (non-blocking): Hey, does this mean that the country filter does not work as intended in the API? Or is it just to say, or is it more of like a sort to sort to the top the country listed? 

I asked just so I can understand more.

cameronapak (approved):
Hey, good work! I approve this, and I think the code looks great. I had one question, and it was mainly just around a comment that was added so I could better understand, but to me that's non-blocking

packages/core/src/__tests__/languages.test.ts
bmanquen: It depends what the intended functionality is supposed to be. Maybe we need to check with someone. The functionality right now is that is just lists the languages that have the countries in them at the top but then it will continue to list languages even if the countries in the filter are not in the countries list on the language until the page_size is returned. I can show you or we can jump on a call for clarity if you want.

PR #90, opened 2026-01-06
YPE-1006: Update AGENTS.md & CLAUDE.md files so that AI agents can better work with this repo

cameronapak:
Because people, including myself, will be using AI agents to interact with and to create and work with this repository, I wanted to make sure that AI has the right guidance that it needs so that when it produces code, it produces code in line with the desires and the standards that we have set forth.

This is the second attempt to make the agent's files better so that AI can be a better assistant and helper as we create and do what God's called us to do

> [!IMPORTANT]
> This does not include a changeset because it doesn't change any features that an end-user interacts with. 

One important piece of information is that the AGENTS.md file is compatible with a ton of AI agents such as people who use OpenCode or if they use Aider or if they use Kilo code or Cursor. However, AGENTS.md is not. They are all incompatible with Claude, which is why I do a symlink and create a CLAUDE.md file for each of the agent files that just points directly to the AGENTS.md file. 

https://agents.md sets the tone for overarching idea and simple application of how AGENTS.md works. 

Jira ticket: https://lifechurch.atlassian.net/browse/YPE-1006

### DIFF
--- a/.changeset/fast-places-juggle.md
+++ b/.changeset/fast-places-juggle.md
@@ -1,0 +1,7 @@
+---
+'@youversion/platform-react-ui': patch
+'@youversion/platform-core': patch
+'@youversion/platform-react-hooks': patch
+---
+
+Refactor verse HTML transformation to support verse-level highlighting. Extract HTML processing logic to `verse-html-utils.ts` with new `wrapVerseContent()` function that wraps verse content in CSS-targetable `<span class="yv-v">` elements. Simplify footnote extraction using wrapped verse structure. Remove CSS rule preventing text wrapping. Add comprehensive test coverage for verse wrapping behavior.
--- a/.gitignore
+++ b/.gitignore
@@ -71,3 +71,5 @@ storybook-static
 !.yarn/sdks
 !.yarn/versions
 .yarn/install-state.gz
+
+.opencode/plans
--- a/packages/ui/src/components/bible-reader.stories.tsx
+++ b/packages/ui/src/components/bible-reader.stories.tsx
@@ -227,7 +227,7 @@ export const FontSizeOutOfRange: Story = {
 
 export const RealAPI: Story = {
   args: {
-    versionId: 111,
+    defaultVersionId: 111,
   },
   parameters: {
     msw: {
--- a/packages/ui/src/components/verse.test.tsx
+++ b/packages/ui/src/components/verse.test.tsx
@@ -319,6 +319,159 @@ describe('Verse.Html - Footnotes', () => {
   });
 });
 
+describe('Verse.Html - Footnote spacing', () => {
+  it('should insert space when footnote is between two words without spacing', async () => {
+    const htmlWithNoSpacing = `
+      <div class="p">
+        <span class="yv-v" v="5"></span><span class="yv-vlbl">5</span>The darkness hasn't overcome<span class="yv-n f"><span class="fr">1:5 </span><span class="ft">Note text</span></span>it.
+      </div>
+    `;
+
+    const { container } = render(<Verse.Html html={htmlWithNoSpacing} renderNotes={true} />);
+
+    await waitFor(() => {
+      const text = container.textContent;
+      expect(text).toContain('overcome it');
+      expect(text).not.toContain('overcomeit');
+    });
+  });
+
+  it('should not insert space when footnote is followed by punctuation', async () => {
+    const htmlWithPunctuation = `
+      <div class="p">
+        <span class="yv-v" v="5"></span><span class="yv-vlbl">5</span>The darkness hasn't overcome<span class="yv-n f"><span class="fr">1:5 </span><span class="ft">Note text</span></span>.
+      </div>
+    `;
+
+    const { container } = render(<Verse.Html html={htmlWithPunctuation} renderNotes={true} />);
+
+    await waitFor(() => {
+      const text = container.textContent;
+      expect(text).toContain('overcome.');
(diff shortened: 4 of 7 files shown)


packages/hooks/AGENTS.md
```diff
@@ -0,0 +1,27 @@
...
+## OVERVIEW
+React integration layer providing 20+ data fetching hooks and 3 core providers.
+
+## STRUCTURE
+- `use*.ts` - Data fetching hooks (useBook, useChapter, usePassage, useVersion, etc.)
+- `context/` - Providers and contexts (separate files, exported via index.ts)
+- `utility/` - Helper functions (useDebounce, extractTextFromHTML, extractVersesFromHTML)
+
+## PUBLIC API
+- 20+ data fetching hooks for YouVersion API resources
```
davidfedor: would it be good to give a few examples on this line? so they can be found or guessed based on the naming convention it would see

packages/ui/AGENTS.md
```diff
@@ -0,0 +1,46 @@
+# AGENTS.md
+
+## OVERVIEW
+Complete UI layer with many Bible components: BibleTextView, VerseOfTheDay, BibleReader, BibleChapterPicker, BibleVersionPicker, YouVersionAuthButton, BibleWidgetView, BibleAppLogoLockup
```
davidfedor: ditto the cross-linking idea

greptile.json
```diff
@@ -0,0 +1,47 @@
+{
+  "commentTypes": ["logic", "syntax", "style"],
+  "instructions": "Enforce project-specific review guidelines - see docs/review-guidelines.md for complete checklist",
+  "customContext": {
+    "rules": [
+      {
+        "scope": ["src/**/*.ts", "src/**/*.tsx", "packages/**/*.ts", "packages/**/*.tsx"],
+        "rule": "All components must use React.forwardRef and accept standard HTML attributes"
```
bmanquen: I don't know if all components need to use `forwardRef`, I think only the components that users might need to interact with child components need refs.

packages/hooks/AGENTS.md
```diff
@@ -0,0 +1,27 @@
...
+- ReaderProvider - Reading session context
+- Utility functions exported from utility/index
+
+## CONVENTIONS
+- Context and Provider in separate files
+- All contexts exported via context/index.ts
+- TypeScript declarations generated separately (no bundling)
+- Build: tsc only
+
+## TESTING
```
bmanquen: Should we also mention testing with mocks and ensuring that the tests are utilizing MSW?

bmanquen (commented):
Great work just some comments that will hopefully help improve a couple areas.

packages/core/AGENTS.md
```diff
@@ -0,0 +1,36 @@
+## OVERVIEW
+Foundation package providing pure TypeScript API clients for YouVersion services with zero React dependencies.
```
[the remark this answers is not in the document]
cameronapak: You ask a good question. When looking into how this works for monorepos, I found this: 

> Instructions from nested AGENTS.md files are combined with parent directories, with more specific instructions taking precedence.
> — [Cursor](https://cursor.com/docs/context/rules#agentsmd)

Because our parent agents file says this:

> ## STRUCTURE
> ```txt
> packages/
>   core/        @youversion/platform-core (API clients, utilities)
>   hooks/       @youversion/platform-react-hooks (React hooks)
>   ui/          @youversion/platform-react-ui (UI components)
> tools/         Shared configs (TS, ESLint)
> ```
> 

Then I think that the agent should have the right context, seeing that this is one of the three packages in the monorepo. 

After being presented with this information, what thoughts do you come to your mind, or do you still have the same question?

greptile.json
cameronapak: Yeah, I agree. This isn't necessary. I think this was for a few components we've created uses this, and not all need it. This was the sort of pattern that I've seen in shadcn-ui components.

packages/ui/AGENTS.md
cameronapak: Please see my response on https://github.com/youversion/platform-sdk-react/pull/90#discussion_r2670149519

packages/ui/AGENTS.md
```diff
@@ -0,0 +1,46 @@
...
+
+## STYLING
+**Auto-injected on import**: `src/index.ts` calls `injectStyles()` on module load
+- CSS embedded as `__YV_STYLES__` constant via tsup define (no separate CSS file)
+- Built Tailwind CSS: `dist/tailwind.css` → injected as JS string at build time
+- Light/dark mode via CSS variables (`[data-yv-sdk]`)
+
+## TESTING
+- Vitest + jsdom for unit tests (`*.test.tsx`)
+- Playwright for Storybook integration tests (`vitest run --project storybook`)
```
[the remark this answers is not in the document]
cameronapak: This has been resolved in my latest set of changes

cameronapak:
I've made some bigger updates to these files based on deep research that [amp code](http://ampcode.com/) did: https://github.com/youversion/platform-sdk-react/pull/90/commits/abf9908fcf84594b96129537c4f0b80bae57532d. It feels like it made the agents' files a lot more clear on what the agent should and shouldn't do.

cameronapak:
Gentle bump requesting review on this. @davidfedor @bmanquen 

background: The context of where I'm coming from is I find that this is incredibly important as I do work, as others do work with AI, and I use AI in my everyday workflow as I'm creating and riffing and digging deeper into code. And so this would mean a lot to have something like this merged in.

PR #94, opened 2026-01-06
YPE-953: fix(ui): fix/refactor the footnotes logic

bmanquen:
- sanitize html before parsing/transforming the string
- use React Portals instead of React Root
- update popover for the footnotes to use new popover structure

### DIFF
--- a/packages/ui/src/components/verse.tsx
+++ b/packages/ui/src/components/verse.tsx
@@ -517,6 +517,7 @@ export const BibleTextView = ({
           lineHeight={lineHeight}
           showVerseNumbers={showVerseNumbers}
           renderNotes={renderNotes}
+          theme={currentTheme}
         />
       </div>
     );
@@ -532,6 +533,7 @@ export const BibleTextView = ({
           lineHeight={lineHeight}
           showVerseNumbers={showVerseNumbers}
           renderNotes={renderNotes}
+          theme={currentTheme}
         />
       </div>
     );
@@ -547,6 +549,7 @@ export const BibleTextView = ({
         showVerseNumbers={showVerseNumbers}
         renderNotes={renderNotes}
         reference={passage?.reference}
+        theme={currentTheme}
       />
     </div>
   );


cameronapak:
CleanShot 2026-01-07 at 10 04 54@2x

---

**suggestion (non-blocking):** There's a way to better align the footnotes so that the line height of the sentence that it's in, or the paragraph that it's in, doesn't get messed up because of the footnote. This isn't something that you need to do in this at all. I just wanted to share that I found a simple way to fix it. On the component that has `data-verse-footnote` I gave it styles of:

```css
{
    display: inline-flex;
    vertical-align: top;
    margin-left: 0.25rem;
}
```

CleanShot 2026-01-07 at 10 02 21@2x

packages/ui/src/components/verse.tsx
```diff
@@ -23,124 +29,166 @@ type ExtractedNotes = {
...
+        if (node.classList.contains('yv-h') || node.closest('.yv-h')) continue;
+        if (node.classList.contains('yv-n') && node.classList.contains('f')) {
+          text += `<sup class="yv:text-muted-foreground">${LETTERS[noteIdx++] || noteIdx}</sup>`;
+        } else if (
+          !node.classList.contains('yv-v') &&
+          !node.classList.contains('yv-vlbl') &&
+          !node.childNodes.length
+        ) {
+          text += node.textContent || '';
+        }
```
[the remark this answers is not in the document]
cameronapak: You're right. I've removed that code in https://github.com/youversion/platform-sdk-react/pull/94/changes/4b7a40e4df2cae10e0f7c0c1bcfa36b1fb3b4d29

packages/ui/src/components/verse.test.tsx
```diff
@@ -298,9 +298,23 @@ describe('Verse.Html - Footnotes', () => {
...
       expect(placeholder).not.toBeNull();
+
       const footnoteElements = container.querySelectorAll('.yv-n.f');
       expect(footnoteElements.length).toBe(0);
     });
+
+    const footnoteButton = container.querySelector('[data-verse-footnote="51"] button');
+    expect(footnoteButton).not.toBeNull();
+
+    fireEvent.click(footnoteButton!);
```
bmanquen: just curious what is the difference between `fireEvent` and `userEvent`?

cameronapak:
@bmanquen I have addressed all the comments and made fixes where needed. I rebased from main so that it has the latest changes, and it should be good for your review again.

packages/ui/package.json
```diff
@@ -49,6 +49,7 @@
     "@radix-ui/react-slot": "1.2.3",
     "@radix-ui/react-tabs": "1.1.13",
     "@radix-ui/react-use-controllable-state": "^1.2.2",
+    "@testing-library/user-event": "^14.6.1",
```
bmanquen: Should this be a dev dependency?

cameronapak: Good catch! Done in https://github.com/youversion/platform-sdk-react/pull/94/commits/ce246d3aed5a5c8d60de10a2d15d2d3957293c05

PR #106, opened 2026-01-16
feat(ui,hooks): add recently used version to the version picker

bmanquen:
- filter out the recently used versions in the useFilterVersions hook to remove them from the main list so they don't show twice.
- add top 3 most recently clicked on bible versions and store them in localStorage.
- allow both lists to be filtered through the searchQuery.

Light mode:
image

Dark mode:
image

### DIFF
--- a/packages/ui/src/components/bible-version-picker.stories.tsx
+++ b/packages/ui/src/components/bible-version-picker.stories.tsx
@@ -4,8 +4,7 @@ import { useState } from 'react';
 import { screen, userEvent, within, expect, waitFor } from 'storybook/test';
 import { BookOpen } from 'lucide-react';
 import { Button } from './ui/button';
-
-const RECENT_VERSIONS_KEY = 'youversion-platform:picker:recent-versions';
+import { RECENT_VERSIONS_KEY } from './bible-version-picker';
 
 type StoredRecentVersion = {
   id: number;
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/bible-version-picker.stories.tsx
```diff
@@ -1,10 +1,22 @@
...
+type StoredRecentVersion = {
+  id: number;
+  title: string;
+  localized_abbreviation: string;
+};
+
+function getStoredRecentVersions(): StoredRecentVersion[] {
+  return JSON.parse(localStorage.getItem(RECENT_VERSIONS_KEY) || '[]') as StoredRecentVersion[];
+}
+
```
cameronapak: **nitpick:** I know that this is weird, but sometimes when I see something like this, I really want to wrap it in a try-catch because I've totally messed with local storage and can see somebody else messing with local storage and then it breaking the app.

packages/ui/src/components/bible-version-picker.stories.tsx
```diff
@@ -1,10 +1,22 @@
 import type { Meta, StoryObj } from '@storybook/react-vite';
 import { BibleVersionPicker, type RootProps } from './bible-version-picker';
 import { useState } from 'react';
-import { screen, userEvent, within, expect } from 'storybook/test';
+import { screen, userEvent, within, expect, waitFor } from 'storybook/test';
 import { BookOpen } from 'lucide-react';
 import { Button } from './ui/button';
 
+const RECENT_VERSIONS_KEY = 'youversion-platform:picker:recent-versions';
```
cameronapak: **nitpick:** Since this is testing the version picker component and we have two of these keys, one in this test file, one in the regular file, can we move that into one file?

PR #121, opened 2026-01-28
YPE-1181: add fields to getLanguages api

bmanquen:
- add fields query param to getLanguages api in core package
- add fields to dependecy array in the useLanguages hook
- add tests to test the new fields query param and page_size=*

### DIFF
--- a/packages/core/src/__tests__/languages.test.ts
+++ b/packages/core/src/__tests__/languages.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, beforeEach } from 'vitest';
+import { describe, it, expect, beforeEach, vi } from 'vitest';
 import { ApiClient } from '../client';
 import { LanguagesClient } from '../languages';
 import { LanguageSchema } from '../schemas';
@@ -61,9 +61,21 @@ describe('LanguagesClient', () => {
       );
     });
 
+    it('should normalize lowercase country codes to uppercase', async () => {
+      const getSpy = vi.spyOn(apiClient, 'get');
+
+      await languagesClient.getLanguages({ country: 'us', page_size: 5 });
+
+      expect(getSpy).toHaveBeenCalledWith(
+        '/v1/languages',
+        expect.objectContaining({ country: 'US' }),
+      );
+      getSpy.mockRestore();
+    });
+
     it('should fetch languages with valid fields filter', async () => {
       const languages = await languagesClient.getLanguages({
-        'fields[]': ['id', 'language', 'script'],
+        fields: ['id', 'language', 'script'],
         page_size: 10,
       });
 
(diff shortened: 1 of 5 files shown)


packages/core/src/languages.ts
cameronapak: I like where you're going. Did you know Zod has power features that would enable this type of validation straight from the schema?

Everything below is what my AI buddy DeepWiki had to say:

--- 

You want to create a single Zod schema that enforces the rule: when page_size is '*', fields[] must have 1-3 items.

Here's how to update the GetLanguagesOptions type with a Zod schema using .refine():

```ts
export const GetLanguagesOptionsSchema = z.object({
  page_size: z.union([z.number().int().positive(), z.literal('*')]).optional(),
  'fields[]': z.array(LanguageSchema.keyof()).optional(),
  country: z.string().optional(),
  page_token: z.string().optional(),
}).refine(
  (data) => {
    if (data.page_size === '*') {
      const fieldsCount = data['fields[]']?.length ?? 0;
      return fieldsCount >= 1 && fieldsCount <= 3;
    }
    return true;
  },
  { message: 'page_size="*" requires 1-3 fields to be specified', path: ['page_size'] }
);
export type GetLanguagesOptions = z.infer<typeof GetLanguagesOptionsSchema>;
```

Then in getLanguages(), you could replace the current piecemeal validation at packages/core/src/languages.ts:57-75 with a single call:

```ts
async getLanguages(options: GetLanguagesOptions = {}): Promise<Collection<Language>> {
  // Single validation call replaces all the individual checks
  GetLanguagesOptionsSchema.parse(options);
  const params: Record<string, string | number | (keyof Language)[]> = {};
  if (options['fields[]']) {
    params['fields[]'] = options['fields[]'];
  }
  if (options.page_size !== undefined) {
    params.page_size = options.page_size;
  }
  // ... rest of param building
}
```

This consolidates all validation into one schema, making it:

1. Declarative - validation rules are defined with the type
2. DRY - no separate validation logic scattered in the function body
3. Type-safe - z.infer derives the TypeScript type automatically

cameronapak (commented):
Hey, thank you for getting this added and working! I know it's odd that we have some dynamic validation based on different fields. 

I put a comment to showcase an alternative solution that uses Zod schemas, and I wanted to get your thoughts on that. It's not a blocking issue or anything, it's just I think that it could be helpful. 

Feel free to push back, and I can press approve

packages/core/src/languages.ts
bmanquen: I did not remember zod could do this either so don't feel bad.

packages/core/src/__tests__/languages.test.ts
```diff
@@ -60,6 +60,109 @@ describe('LanguagesClient', () => {
         'Country code must be a 2-character ISO 3166-1 alpha-2 code',
       );
     });
+
+    it('should normalize lowercase country codes to uppercase', async () => {
+      const languages = await languagesClient.getLanguages({ country: 'us', page_size: 5 });
+      expect(languages.data.length).toBeGreaterThan(0);
+    });
```
cameronapak: [question](https://conventionalcomments.org/): Hey I need a little bit help understanding this test. The title of the test seems different than what's inside of the test and maybe I just don't understand it. Does the title match the test code?

packages/core/src/__tests__/languages.test.ts
```diff
@@ -60,6 +60,109 @@ describe('LanguagesClient', () => {
...
+
+    it('should fetch languages with valid fields filter', async () => {
+      const languages = await languagesClient.getLanguages({
+        fields: ['id', 'language', 'script'],
+        page_size: 10,
+      });
+
+      expect(languages.data).toHaveLength(10);
+    });
```
cameronapak: [praise](https://conventionalcomments.org/): this looks great. Simple

packages/core/src/__tests__/languages.test.ts
bmanquen: Great question I blame Claude! I think that if we assert that the country code gets uppercased it would be more clear.

Is this what you are expecting?

packages/core/src/__tests__/languages.test.ts
```diff
@@ -60,6 +60,109 @@ describe('LanguagesClient', () => {
...
+      ).rejects.toThrow('page_size="*" requires 1-3 fields to be specified');
+    });
+
+    it('should allow page_size="*" with exactly 1 field', async () => {
+      const languages = await languagesClient.getLanguages({
+        fields: ['id'],
+        page_size: '*',
+      });
+      expect(languages.data.length).toBeGreaterThan(0);
+    });
```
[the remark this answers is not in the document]
bmanquen: There should also be a test to validate your question that you had.

cameronapak (approved):
passes with flying colors. approved!

PR #122, opened 2026-01-29
YPE-1154: Add Test For ReaderProvider Hooks

jaredhightower-youversion:
# Primary Changes
- Added test suite for ReaderProvider context functionality
- Test version, book, chapter, and verse navigation state management
- Verify navigation state persistence across updates
- Fix null verse rendering to use an empty string instead of 'null' literal

# Additional Info
Screenshot 2026-01-30 at 9 43 18 AM

### DIFF
--- a/packages/hooks/src/types/bibles.ts
+++ b/packages/hooks/src/types/bibles.ts
@@ -1,8 +1,0 @@
-import type { BibleBook, BibleChapter, BibleVerse, BibleVersion } from '@youversion/platform-core';
-
-export interface ReaderProviderOptions {
-  currentVersion: BibleVersion;
-  currentBook: BibleBook;
-  currentChapter: BibleChapter;
-  currentVerse: BibleVerse | null;
-}


packages/hooks/src/__tests__/utils/test-utils.tsx
```diff
@@ -41,3 +50,29 @@ export function TestAuthChild({ onRender }: { onRender?: (data: any) => void }):
...
   );
 }
+
+/**
+ * Renders a component wrapped in ReaderProvider for testing
+ * @param ui Component to render
+ * @param options Optional provider props to override defaults
+ */
+export const renderWithReaderProvider = (
+  ui: React.ReactElement,
```
bmanquen: Do we want to use the wider `React.ReactNode` type?

packages/hooks/src/types/bibles.ts
```diff
@@ -0,0 +1,8 @@
+import type { BibleBook, BibleChapter, BibleVerse, BibleVersion } from '@youversion/platform-core';
+
+export interface ReaderProviderOptions {
+  currentVersion: BibleVersion;
+  currentBook: BibleBook;
+  currentChapter: BibleChapter;
+  currentVerse: BibleVerse | null;
+}
```
bmanquen: Lets make this a type and just put it in the `ReaderProvider.tsx` file if we need this.

bmanquen (changes requested):
Just a few comments/changes, feel free to push back on any comments if you disagree.

packages/hooks/src/__tests__/utils/test-utils.tsx
cameronapak: In order to provide feedback that's faster towards resolution can you consider adoptin https://conventionalcomments.org

In my head I can try to reason that the question is a nitpick/question that's non-blocking. Is that what you would also say @bmanquen?

packages/hooks/src/types/bibles.ts
cameronapak: [thought](https://conventionalcomments.org/): I'm aligned with Brenden's suggestion

packages/hooks/src/__tests__/utils/test-utils.tsx
jaredhightower-youversion: Good catch, this should have been `React.ReactNode`

cameronapak (changes requested):
Great work, Jared! I'm excited that you guys got in and got running and were able to contribute pretty quickly. I think the code looks good. 

There's a couple comments that I had, nothing big. Then I'm going to mark this as requesting changes because Brenden had a suggestion that I align with, just moving a type into a pre-existing file, not into its own file.

packages/hooks/src/types/bibles.ts
jaredhightower-youversion: Looks like Cam already added those types in that file 😁

packages/hooks/src/context/ReaderProvider.test.tsx
cameronapak: After thinking... I have a suggestion. 

> [!NOTE]
> **I had AI help me format this well, but I reviewed every line and stand by the suggestion:**

---

[suggestion](https://conventionalcomments.org/): Hey! This could be simplified using renderHook instead of the DOM-based approach. Since we're testing `useReaderContext` (a hook), we can test it in isolation without needing a `TestChild` component:

```ts
import { renderHook, act } from '@testing-library/react';
const wrapper = ({ children }) => (
  <ReaderProvider
    currentVersion={mockVersion}
    currentBook={mockBook}
    currentChapter={mockChapter}
    currentVerse={mockVerse}
  >
    {children}
  </ReaderProvider>
);
it('should update version when setVersion is called', () => {
  const { result } = renderHook(() => useReaderContext(), { wrapper });
  
  act(() => {
    result.current.setVersion(mockVersion2);
  });
  
  expect(result.current.currentVersion).toEqual(mockVersion2);
});
```

This removes the need for:
- The `TestChild` component (lines 80-115)
- `data-testid` attributes and DOM queries
- Button clicks to trigger setters
- JSON serialization/parsing to check state

The tests become more direct — we access result.current to read state and call setters via act(). Same coverage, less code.

packages/hooks/src/types/bibles.ts
jaredhightower-youversion: thought: Yes, I think so. This was mostly used by the `renderWithReaderProvider` test-util, which was removed, but we can keep it if you think it will add value or be used in the future. My vote is that we remove it.

bmanquen: Lets go ahead and remove it, I think we all agree with that for now.

PR #152, opened 2026-02-17
Refactor verse rendering and footnote extraction

jaredhightower-youversion:
## Summary
This change refactors verse footnote handling so markers are rendered inline at their true text position, while footnote popovers still anchor correctly and spacing remains intact across Bible versions.

### What changed
Reworked footnote extraction to traverse wrapped verse content with a scoped TreeWalker.
Added efficient verse grouping maps for footnotes and wrappers to reduce repeated DOM queries.
Updated marker generation to support multi-letter sequences (a…z, aa, ab, etc.).
Attached the portal anchor to the first inline footnote in each verse, with a fallback placeholder when needed.
Preserved text spacing after marker insertion to prevent word concatenation edge cases.
Kept gutted footnote elements in the DOM as empty shells (instead of removing them) to maintain natural text boundaries.
In verse rendering, removed debug logging and ensured the content container renders before portal targets are resolved.

### Why
Footnote markers now reflect the actual inline location users expect.
Popover placement is more reliable for verses with multiple notes or unusual markup.
The refactor improves readability, correctness, and performance of note extraction.

### Impact
No API surface changes.
UX improvement for Bible text footnote display and anchor behavior.
Better resilience with malformed or tightly spaced source HTML.

## Additional Info (Videos, Screenshot, etc)
https://github.com/user-attachments/assets/5bf7ee71-7781-4bf7-b1ee-11bc8b9fd58e

### DIFF
--- a/.changeset/great-showers-fry.md
+++ b/.changeset/great-showers-fry.md
@@ -1,0 +1,14 @@
+---
+'@youversion/platform-react-ui': patch
+'@youversion/platform-core': patch
+'@youversion/platform-react-hooks': patch
+---
+
+Refactor verse footnote extraction and rendering for clarity and correctness
+
+- Replace TreeWalker-based footnote extraction with clone-and-transform approach
+- Move HTML transformation pipeline into `verse-html-utils.ts` as `transformBibleHtml`
+- Fix space insertion between element siblings when footnotes are removed
+- Fix footnote marker/label mismatch for verses with >26 footnotes
+- Simplify `BibleTextHtml` hooks and use React `onClick` instead of manual event listeners
+- Use `useMemo` for synchronous HTML transformation instead of `useEffect` + `useState`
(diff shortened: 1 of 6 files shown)


bmanquen:
There is still only one footnote popover trigger per verse. It seems to be placing it at the first instance of a footnote.

Current Implementation (This branch)
image
 
Current Android implementation
image

I believe it needs to follow the Android and Swift functionality where we show a footnote icon as the popover trigger inline where the note is. If you have questions or need clarity please feel free to reach out.

cameronapak (changes requested):
This looks good as long as it works. I know it's been a long journey on this one. You're doing great. 

I have one request. Can you create a test, which can be in `verse.stories.tsx` that validates multiple footnotes appearing in a single verse, really what brenden pointed out earlier in his review. Doesn't have to be anything fancy. Just simple do the two footnotes exist in the right spot? If so, great.

cameronapak:
@greptileai, thanks for your review. You've seen how this problem was approached. How would you approach this problem?

Kyleasmth (dismissed):

cameronapak (approved):
Greptile's Confidence Score is a 5/5, and so is mine. Approved

PR #203, opened 2026-03-26
fix(ui): show specific Bible passage error messages

davidfedor:
Map BibleTextView errors to clearer user-facing messages for missing passages, missing or invalid app keys, forbidden access, rate limits, unreachable server/timeouts, and 5xx responses.

Extract the error-message mapping into a dedicated UI helper and extend component tests to cover the new cases.


cameronapak (dismissed):
This looks good to me. Thank you for making it where we have better, more specific error copy! Can you run `pnpm changeset` at the project's root to bump all of the packages, so that the changes propogate into the next version? _(we bump them all at once for them to stay on the same version across the board for react sdk packages)_

davidfedor:
@cameronapak changeset added; care to re-review?

cameronapak (approved):

cameronapak:
Approved!

PR #210, opened 2026-04-08
chore: implement i18next

camrun91:
This PR puts i18next into place and sets up a single translation string to show how it will work. 

The current implementation is to mark i18next as a peer dependency and make it required. This is to prevent conflicts with any implementation apps having their own i18next implementation.

<h3>Greptile Summary</h3>

This PR wires up i18next to the `@youversion/platform-react-ui` package using an isolated instance (`createInstance()`) so the SDK's translations can never collide with a host app's own i18next configuration. The only translated surface right now is the "Verse of The Day" label in `VerseOfTheDay`. Previous concerns (unified versioning, silent init failure, SSR-breaking `LanguageDetector`, and the `optional: true` / static-import contradiction) have all been addressed in this revision.

Safe to merge — all previously raised blockers are resolved; only two minor P2 suggestions remain.

Unified versioning is now correct (all three packages bumped), the LanguageDetector SSR risk is gone, init errors are observable via .catch(), and the isolated-instance design is intentional and coherent. The remaining findings are P2: a changeset description that mischaracterises the deps as 'optional peer dependencies' and a missing explicit lng that may produce a dev-mode warning. Neither blocks correctness.

The changeset description in .changeset/add-i18next-support.md should be corrected before merge to avoid misleading changelog consumers.

<details open><summary><h3>Vulnerabilities</h3></summary>

No security concerns identified. i18next resources are in-memory constants, no user-supplied input reaches the translation layer, and `escapeValue: false` is safe because React's JSX already escapes all output.
</details>

| Filename | Overview |
|----------|----------|
| .changeset/add-i18next-support.md | Changeset now correctly bumps all three packages (unified versioning addressed), but the description inaccurately calls i18next/react-i18next 'optional peer dependencies' when they are regular bundled deps. |
| packages/ui/src/i18n/index.ts | Isolated i18next instance via createInstance() with .catch() error handler; LanguageDetector removed (SSR-safe); no explicit lng may produce a dev-mode warning. |
| packages/ui/package.json | i18next and react-i18next added as regular dependencies (bundled isolated instance approach, consistent with tsup config); no peerDependenciesMeta or optional markers. |
| packages/ui/src/components/verse-of-the-day.tsx | useTranslation correctly pinned to SDK's isolated i18n instance via { i18n } option; replaces hardcoded string with t('verseOfTheDay'). |
| packages/ui/src/i18n/locales/en.json | Single translation key for 'Verse of The Day' in English locale; clean baseline for future strings. |
| .gitignore | Added .omc/ to the ignore list; straightforward tooling artifact exclusion. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A["packages/ui/src/i18n/index.ts\n(module load)"] -->|"createInstance()"| B["Isolated i18n instance"]
    B -->|"use(initReactI18next)\n.init({ resources })"| C["In-memory EN resources loaded"]
    C --> D["i18n instance exported"]
    D --> E["verse-of-the-day.tsx\nimport i18n from '@/i18n'"]
    E -->|"useTranslation(undefined, { i18n })"| F["t() bound to SDK instance"]
    F -->|"t('verseOfTheDay')"| G["'Verse of The Day'"]
    H["Host app's i18next instance"] -.->|"Completely isolated\nno interference"| B
```

<details><summary>Prompt To Fix All With AI</summary>

`````markdown
This is a comment left during a code review.
Path: .changeset/add-i18next-support.md
Line: 7

Comment:
**Changeset description doesn't match the implementation**

The description states "i18next and react-i18next are optional peer dependencies", but in `packages/ui/package.json` both are listed under regular `dependencies`, not `peerDependencies`. They will be **bundled** into the SDK output (they're also absent from tsup's `external` array). A consumer who reads the changelog may incorrectly assume they don't need to install either package themselves, or that they can omit them without consequence.

The description should reflect the actual approach — isolated bundled instance — so consumers understand the bundle size tradeoff and don't expect to supply their own copies.

How can I resolve this? If you propose a fix, please make it concise.

---

This is a comment left during a code review.
Path: packages/ui/src/i18n/index.ts
Line: 15-22

Comment:
**No explicit `lng` — i18next may emit a console warning in dev**

Without an explicit `lng` and no language-detector plugin, i18next v26 internally sets the language to `undefined` and falls back to `fallbackLng`. In development mode this can produce a warning like "i18next: init: no language detected". Since the SDK only ships English strings, setting `lng: 'en'` explicitly makes the intent clear, silences any warning, and prevents any edge-case where a future i18next version changes detection behaviour.

```suggestion
  .init({
    resources,
    lng: 'en',
    defaultNS,
    fallbackLng: 'en',
    interpolation: {
      escapeValue: false, // React already escapes
    },
  })
```

How can I resolve this? If you propose a fix, please make it concise.
`````

</details>

<sub>Reviews (10): Last reviewed commit: ["fix: internalize i18next"](https://github.com/youversion/platform-sdk-react/commit/c524bb191e64268382727e9619895829b2870832) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=27742681)</sub>

**Context used:**

- Rule used - Unified versioning: All packages must maintain exa... ([source](https://app.greptile.com/review/custom-context?memory=auto-2977a6cb))

---

<a href="https://app.devin.ai/review/youversion/platform-sdk-react/pull/210" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

### CODE REVIEW

File: packages/ui/src/components/verse-of-the-day.tsx [hunk 1]
```diff
@@ -6,6 +6,8 @@ import {
   useVersion,
 } from '@youversion/platform-react-hooks';
 import React from 'react';
+import { useTranslation } from 'react-i18next';
+import i18n from '@/i18n';
```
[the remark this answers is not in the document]
camrun91: @greptile-apps these are already present in the external array

packages/ui/package.json
```diff
@@ -53,6 +53,9 @@
     "@youversion/platform-react-hooks": "workspace:*",
     "class-variance-authority": "0.7.1",
     "clsx": "2.1.1",
+    "i18next": "^26.0.4",
+    "i18next-browser-languagedetector": "^8.2.1",
+    "react-i18next": "^17.0.2",
```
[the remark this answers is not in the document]
camrun91: @bmanquen  Here is the note on peerDependencies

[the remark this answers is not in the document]
bmanquen: I have actually read that you do not want to do this, instead you want to scope the translations to their own instance of i18next.

packages/ui/src/components/verse-of-the-day.tsx
```diff
@@ -89,6 +91,7 @@ export function VerseOfTheDay({
   showBibleAppAttribution = true,
   size = 'default',
 }: VerseOfTheDayProps): React.ReactElement {
+  const { t } = useTranslation(undefined, { i18n });
```
bmanquen, line 94: Is this just going to use the default namespace and then override the i18n instance with our instance?

bmanquen (approved):
Looks good just one comment for clarification/learning.

packages/ui/src/components/verse-of-the-day.tsx
camrun91: yep that is exactly what it is doing.

camrun91:
@bmanquen I am going to wait and let @cameronapak take a look at this too and make sure it tracks with what he thinks is best.

packages/ui/src/i18n/index.ts
```diff
@@ -0,0 +1,27 @@
...
+const defaultNS = 'translation';
+
+export const resources = {
+  en: { [defaultNS]: en },
+} as const;
+
+const i18n: I18nInstance = i18next.createInstance();
+
+i18n
+  .use(initReactI18next)
```
cameronapak, line 14: 🔴 **`.use(initReactI18next)` mutates a global singleton, potentially breaking consumer i18n setups**

The SDK creates an isolated i18next instance via `createInstance()` but then calls `.use(initReactI18next)` on it. When `i18n.init()` runs, `initReactI18next.init(sdkInstance)` is invoked, which internally calls `setI18n(sdkInstance)` — storing the SDK's instance in a **module-level variable** inside `react-i18next`. If a consumer also uses `react-i18next` with their own i18n instance (and shares the same `react-i18next` module copy), the SDK's `init()` silently overwrites the consumer's default i18n reference. Any consumer component that calls `useTranslation()` without an `<I18nextProvider>` would then resolve to the SDK's instance (with only one `verseOfTheDay` key) instead of their own.

Critically, this side effect runs at **module load time** for every consumer of `@youversion/platform-react-ui` — not just those using `VerseOfTheDay` — because tsup is configured with `splitting: false`, so the i18n module is eagerly evaluated when any export is imported.

The `.use(initReactI18next)` call is also **unnecessary**: the component at `packages/ui/src/components/verse-of-the-day.tsx:94` already passes the explicit instance via `useTranslation(undefined, { i18n })`, which bypasses the default instance lookup entirely. The fix is to remove `.use(initReactI18next)` and just call `i18n.init({...})` directly.

Note: the changeset description at `.changeset/add-i18next-support.md:7` claims "no global singleton mutation," which is factually incorrect.

<details>
<summary>Prompt for agents</summary>

```
In packages/ui/src/i18n/index.ts, the call to .use(initReactI18next) on line 14 should be removed. The initReactI18next plugin is a singleton that stores a reference to the i18n instance it was initialized with. When the SDK calls .use(initReactI18next).init(...), it overwrites whatever i18n instance the consumer may have previously registered via their own initReactI18next setup. This can silently break consumer apps that use react-i18next.

The fix is straightforward: remove .use(initReactI18next) from the chain, and also remove the import of initReactI18next from react-i18next. The SDK component already passes { i18n } explicitly to useTranslation in verse-of-the-day.tsx:94, so the initReactI18next plugin is not needed. The i18n setup should become:

i18n.init({ resources, defaultNS, fallbackLng: 'en', interpolation: { escapeValue: false } }).catch(...);

Note that useTranslation from react-i18next can still be used in the component — it accepts the i18n instance directly via the options parameter. The import of useTranslation in verse-of-the-day.tsx does not need to change.

Also update the changeset description in .changeset/add-i18next-support.md which currently says 'no global singleton mutation' — this claim would only become true after the fix.
```

</details>


<a href="https://app.devin.ai/review/youversion/platform-sdk-react/pull/210" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>


---
*Was this helpful? React with 👍 or 👎 to provide feedback.*

cameronapak (commented):
**Devin Review** found 1 potential issue.

View 4 additional findings in Devin Review.


<a href="https://app.devin.ai/review/youversion/platform-sdk-react/pull/210" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

packages/ui/src/i18n/index.ts
cameronapak: Nope... this looks like it's standard practice. See https://www.locize.com/blog/react-i18next/

cameronapak:
This looks great. No qualms. Great work, @camrun91!

PR #216, opened 2026-04-20
feat(core): auto-transform Bible HTML in getPassage

cameronapak:
## Summary

Auto-transforms Bible HTML inside `getPassage` so consumers never need to call `transformBibleHtml` manually. Uses native `DOMParser` in the browser and dynamic `import('jsdom')` on the server. Added `data-yv-transformed` idempotency marker so double-transforms are a no-op.

**Runtime DOM choice:** optional peer is `jsdom` (not linkedom). No workerd-specific path or CI smoke — server consumers that need auto-transform install `jsdom`, or pass `transform: false` to skip.

`Verse.Html` retains its `transformBibleHtml` call as defense-in-depth — the idempotency marker makes it a no-op for HTML that already went through `getPassage`.

**Node floor:** `jsdom@28.1.0` (devDep / peer `^24 || ^28`) declares `^20.19.0 || ^22.12.0 || >=24.0.0`, which covers the repo `engines.node >=22.13` floor (Decision 3).

Context: [Why transformBibleHtml Exists — And Where It May Not Be Needed](https://www.notion.so/yvproduct/react-sdk-Why-transformBibleHtml-Exists-And-Where-It-May-Not-Be-Needed-333f1f2d1b928000b8f6f48c465457ef)

## Test plan

- [x] Verify `getPassage` with `format: 'html'` returns transformed content (`data-yv-transformed` present)
- [x] Verify `getPassage` with `format: 'text'` returns raw content (no transformation)
- [x] Verify double-transform produces identical output (idempotency)
- [x] Verify footnotes are extracted into `data-verse-footnote` attributes
- [x] Verify `Verse.Html` still sanitizes raw HTML passed directly (XSS protection)

<h3>Greptile Summary</h3>

The PR makes `getPassage` automatically sanitize and structurally transform HTML while retaining explicit opt-out and runtime-specific DOM behavior.
- Adds browser-native and optional server-side `jsdom` adapters.
- Makes transformation idempotent across nested markers and multi-root fragments.
- Forwards the transformation option through `usePassage`.
- Updates UI integration, styling, tests, package metadata, and build configuration.

The PR appears safe to merge.

No blocking failure remains, and the previously reported marker, dependency-error, type-version, sibling-root, and hook-forwarding issues are fixed or invalid at the current head.

| Filename | Overview |
|----------|----------|
| packages/core/src/bible-html-transformer.ts | Adds sanitized, multi-root-aware idempotency handling and marks every transformed top-level element. |
| packages/core/src/bible.ts | Automatically transforms HTML passages using browser DOMParser or optional server-side jsdom, with an explicit opt-out. |
| packages/hooks/src/usePassage.ts | Exposes, forwards, and correctly keys fetching on the passage transformation option. |
| packages/core/src/bible-html-transformer-server.ts | Replaces the server transformer’s linkedom adapter with jsdom. |
| packages/core/package.json | Aligns jsdom runtime and type versions, declares it as an optional peer, and prevents browser resolution. |
| packages/ui/src/components/verse.tsx | Retains client-side transformation as an idempotent rendering safety layer. |

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
  participant Consumer
  participant Hook as usePassage
  participant Core as BibleClient.getPassage
  participant API as YouVersion API
  participant DOM as DOMParser or jsdom
  Consumer->>Hook: Request passage with transform option
  Hook->>Core: Forward request and transform option
  Core->>API: Fetch passage
  API-->>Core: Passage content
  alt HTML and transform enabled
    Core->>DOM: Parse content
    DOM-->>Core: Document
    Core->>Core: Sanitize and structurally transform
  end
  Core-->>Hook: Passage
  Hook-->>Consumer: Passage state
```

<sub>Reviews (18): Last reviewed commit: ["Merge branch &#39;main&#39; into transform-bible..."](https://github.com/youversion/platform-sdk-react/commit/1948fac2d0c91d6d3824871792fac52f858a8fc7) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=29027239)</sub>

- Knowledge Base — [packages/hooks — React hooks for the YouVersion Platform](https://app.greptile.com/youversion/-/custom-context/knowledge-base/youversion/platform-sdk-react/-/docs/hooks-package.md)
</details>


packages/core/src/bible-html-transformer.test.ts
```diff
@@ -308,7 +308,7 @@ describe('transformBibleHtml - sanitization', () => {
     const result = transformBibleHtml(html, createAdapters());
 
     expect(result.html).not.toContain('onclick');
-    expect(result.html).toContain('<p>');
+    expect(result.html).toContain('<p');
```
cameronapak: note: if you're wondering why this tag is _seemingly_ cut off, it's because the tags would contain a data attribute in this new PR, which would then make it where the tag is something like `<p data-yv-attribute>` versus `<p>` standalone

packages/core/src/bible.ts
```diff
@@ -13,6 +14,33 @@ import type {
...
+      serializeHtml: (doc) => doc.body.innerHTML,
+    };
+  }
+  let linkedom;
+  try {
+    linkedom = await import('linkedom');
+  } catch {
+    throw new Error(
+      'Server-side HTML transformation requires "linkedom". ' +
+        'Install it as a dependency or pass format: "text" to skip transformation.',
```
davidfedor: This might be better if there was a supported way to get raw html (untransformed), for people who don't want to import linkedom or who (for whatever reason) want the original data. How about a new format option, "rawhtml" or something like that?

davidfedor: (I'm writing this here because this error path is not something a builder will probably be excited to be in. The fix would mostly be elsewhere.)

davidfedor: ... or add another parameter so that the format can stay "html". That feels like a better idea to me.

davidfedor (commented):
Big picture: I love the idea of being helpful, without requiring the developer to have to make another call. My comments and questions are around whether this is the best way to do that. (Maybe it is! I'm not sure yet.)
I notice this would be blurring the lines between Core being merely an API helper-layer, but now it would be doing some of the prep-work of the UI (visualization layer). So at the least having that be optional seems wise.
I'm wondering if that parameter should default to do the transformation, or not... or whether we need to force the dev to make a choice (to attempt to force them to make an informed choice).

davidfedor:
(FYI I've asked for thoughts from Bryson H; not sure if he's got cycles to contribute or not)

packages/core/src/bible.ts
cameronapak: I like what you're processing. I've added a new commit to have the escape hatch allowing users to intentionally seek raw html versus transformed: https://github.com/youversion/platform-sdk-react/pull/216/commits/aece62a6ad8ebccbc376c0d7b47093f360827a28

This PR is ready for re-review and re-consideration @davidfedor

arinthros: A few things:

- I'm curious, why `linkedom` as opposed to a more widely supported library like `jsdom`? Is there any risk of supply chain pollution with the newer library?
- Have the docs been updated to reflect the need for a third party dependency?
- Can the dependency be added as an optional peer dependency so it shows up in install logs?

arinthros: I ask as someone who is doing RSC data loading, and will need to have this work server-side :)

packages/core/src/bible-html-transformer.test.ts
davidfedor: can this `expect()` call do regular expressions? we don't have `pre` tags yet, that I know of, but still... it'd be great to tighten this up if that's not too difficult.

packages/core/src/bible.ts
cameronapak: Hey Bryson (@arinthros)! 

Based on your comment, I've moved from `linkedom` to `jsdom`. 

The YV dev docs will need to be updated immediatley after this (I will do that)

The dep has been added as an optional peer dep, so it can show up in install logs.

Anything else blocking this?

cameronapak: @arinthros following up ^

arinthros:
@davidfedor @cameronapak this looks good to me, I just don't see an "approve" button in my UI. Approved by me!

cameronapak:
Some added context on this ticket is: [Why transformBibleHtml Exists — And Where It May Not Be Needed](https://www.notion.so/yvproduct/react-sdk-Why-transformBibleHtml-Exists-And-Where-It-May-Not-Be-Needed-333f1f2d1b928000b8f6f48c465457ef)

If we can get this merged in, then I can write better docs for helping people use our HTML and styles without them having to manually `transformBibleHtml` on their end

packages/core/src/bible-html-transformer.test.ts
davidfedor: ping on this task

cameronapak:
Adding some clarity. David sent me a DM on Teams the other day saying this:

> Cam, at your leisure, can you rebase [Auto-transform the Bible HTML from getPassage so the consumer doesn't have extra steps by cameronapa…](https://github.com/youversion/platform-sdk-react/pull/216) (or merge, whichever) and have greptile re-review it? I finally had time to look at it again and consider things and I think it's a step in the right direction.
> (From a strategic point of view I'd like to make it easier for agents to create pages using YVP, and supporting non-React web paths is a component of that.)

packages/core/src/bible-html-transformer.test.ts
cameronapak: Tightened in `ad51471` to a tag-boundary regex:

```ts
expect(result.html).toMatch(/<p(?:\s[^>]*)?>Click me<\/p>/);
```

So a false positive like `<pre` no longer matches.

packages/core/src/bible-html-transformer.ts
```diff
@@ -339,6 +343,11 @@ export function transformBibleHtml(
   const doc = options.parseHtml(html);
 
   sanitizeBibleHtmlDocument(doc);
+
+  // Already transformed — skip structural transforms
+  if (doc.querySelector(`[${TRANSFORMED_ATTR}]`)) {
+    return { html: options.serializeHtml(doc) };
```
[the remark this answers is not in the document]
cameronapak: Fixed in `ad51471`: idempotency now checks only the transform root (`body.firstElementChild ?? body`) for `data-yv-transformed`, not `querySelector` anywhere in the document. Nested/untrusted markers no longer skip verse wrap or footnote extraction. Added a regression test for that case.

packages/core/src/bible.ts
cameronapak: Already shipped earlier as `transform: false` on `getPassage` — resolving this thread as addressed.

packages/core/src/bible-html-transformer.ts
```diff
@@ -339,6 +341,15 @@ export function transformBibleHtml(
...
 
   sanitizeBibleHtmlDocument(doc);
+
+  // Only trust the marker on the transform root. A nested/untrusted
+  // data-yv-transformed (preserved by data-* allowlist) must not skip
+  // verse wrapping or footnote extraction.
+  const root = doc.body?.firstElementChild ?? doc.body;
+  if (root?.hasAttribute(TRANSFORMED_ATTR)) {
+    return { html: options.serializeHtml(doc) };
+  }
```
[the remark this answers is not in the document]
cameronapak: Valid — fixed in `7206abc`.

The guard now trusts the marker only when **every** top-level element carries it, and `transformBibleHtml` marks every top-level element instead of just the first. A fragment mixing transformed and raw roots now runs the full structural transform.

Two regression tests added: raw sibling after a marked root still gets verse wrapping and footnote extraction, and a multi-root fragment stays idempotent across two passes.

packages/core/src/bible.ts
```diff
@@ -13,6 +14,36 @@ import type {
...
+    // Literal dynamic import is fine in Node. Client bundlers must not pull
+    // jsdom into browser graphs — see package.json "browser": { "jsdom": false }.
+    jsdom = await import('jsdom');
+  } catch (err) {
+    const detail = err instanceof Error ? err.message : String(err);
+    throw new Error(
+      'Server-side HTML transformation requires "jsdom". ' +
+        'Install it as a dependency or pass transform: false to skip transformation. ' +
+        `Original error: ${detail}`,
+      { cause: err },
```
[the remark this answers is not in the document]
cameronapak: Partly right, and fixed in `747c905`.

The API gap was real: `getPassage` has a `transform` escape hatch and `usePassage` silently dropped it. `usePassage` now accepts `transform` and forwards it, so a caller can opt out of HTML transformation through the hook.

One correction on the mechanism: the described SSR failure is not reachable through React SSR. `useApiData` fetches inside `useEffect`, and effects do not run during server rendering, so `usePassage` always fetches in the browser where native `DOMParser` exists — jsdom is never loaded on that path. The gap that did matter was ergonomic: non-DOM Node or test environments, and callers who simply want the raw HTML.

jhampton:
@davidfedor - the Idempotency guarantee you mentioned was already implemented and Greptile shows 5/5 - is there another place I'm not seeing?

cameronapak:
Hey @jhampton! David shared the other day in a DM:

> Different topic: to test/validate/demo [feat(core): auto-transform Bible HTML in getPassage by cameronapak · Pull Request #216 · youversion/…](https://github.com/youversion/platform-sdk-react/pull/216) what if you made a very simple html page which makes the API call, uses this new auto-transform, and puts the html into a div on the page? Simplest (but still rational and useful) example, and that could become material for a Guide page.

I responded with: 

> I like your idea. This would have to be local since we do not have a way to do interim/test npm package publishing. So, the testing artifact would have to be local. Would you want to see a video with the demo code sample? 

He responded:

> Whatever's convenient for you!

So... I need to make a testing artifact and share the results with David. However, I've been focused on React Native Expo SDK highlights, so balancing this plate spinning while other bigger plates are spinning is what's causing my delays.

cameronapak:
Okay, I'm going to share two demos. One will be in React, the other one will be just vanilla TypeScript. The idea is getting the Bible text and being able to display it without having to do all the mutation transform HTML. At least the user does not have to do that. It happens automatically. 
 
1. [React example](https://video.cpak.me/s/nqgwwhy0ghgvxp5)
2. [Vanilla TS example](https://video.cpak.me/s/weez9xkn882wac6)

@davidfedor, these are examples of this working and showing that the user does not have to do any transform HTML function to get bible text to be displayed properly

cc: @jhampton

davidfedor (approved):

davidfedor:
Thanks, Cam!

PR #264, opened 2026-06-11
YPE-1565: Show error when YouVersionProvider has no appKey

cameronapak:
image

Empty/missing `appKey` rendered a blank page. Now: UI provider shows a styled "Missing app key" message; hooks provider throws.

- Styled `MissingAppKey` panel + Storybook story
- i18n (en/fr/es), example cleanup, unit + integration tests, changeset

[YPE-1565](https://lifechurch.atlassian.net/browse/YPE-1565)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[YPE-1565]: https://lifechurch.atlassian.net/browse/YPE-1565?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

<h3>Greptile Summary</h3>

This PR surfaces a clear error when `YouVersionProvider` is given a missing or empty `appKey`, replacing the previous blank-page failure mode. The UI package renders a styled `MissingAppKey` panel with i18n support, while the hooks package throws a descriptive error (for hooks-only consumers who own their own error boundary).

- **Hooks provider** is split into a thin `YouVersionProvider` wrapper (validates `appKey`, throws if invalid) and an inner `YouVersionProviderInner` that owns all hook calls, keeping hook order consistent across renders regardless of key validity.
- **UI provider** detects the missing key before calling the base provider, logs to `console.error` from an effect (avoiding Strict Mode double-calls), and renders the `MissingAppKey` panel with `YvStyles` scoped styles and proper `role="alert"` semantics.
- Tests, Storybook story, i18n translations (en/fr/es), changeset, and example cleanup are all included.

Safe to merge. The two-layer guard (styled panel in UI, throw in hooks) is correctly implemented, hook order is preserved by the wrapper/inner split, and all three invalid-key variants are covered by tests.

The wrapper/inner split cleanly separates the validation logic from hook-bearing code, the UI provider calls hooks unconditionally before any conditional return, and the console.error is deferred to an effect to avoid Strict Mode duplication. No logic defects were found in the changed paths.

No files require special attention.

| Filename | Overview |
|----------|----------|
| packages/hooks/src/context/YouVersionProvider.tsx | Split into a thin validation wrapper and an inner hook-bearing component; the throw guard lives before any hook calls, correctly preserving hook order on every render path. |
| packages/ui/src/components/YouVersionProvider.tsx | Adds `missingAppKey` guard with a `useEffect`-based console.error (avoids Strict Mode duplication) and renders `MissingAppKey` panel with `resolveTheme` before ever reaching the base provider. |
| packages/ui/src/components/missing-app-key.tsx | New component with correct `role="alert"` semantics (no overriding `aria-live`), i18n via react-i18next, and scoped Tailwind classes. Uses `data-yv-theme` for theming. |
| packages/hooks/src/context/YouVersionProvider.test.tsx | Adds `it.each` tests covering undefined, empty string, and whitespace-only appKey cases; each verifies the correct throw message. |
| packages/ui/src/components/YouVersionProvider.test.tsx | Tests verify `role="alert"` presence, error heading text, child suppression, and `console.error` call with `appKey` substring for all three invalid-key variants. |
| packages/ui/src/components/missing-app-key.stories.tsx | Adds Light, Dark, and integration stories with a play function asserting `role="alert"`, heading, and body text. |
| examples/vite-react/src/ThemedApp.tsx | Removes the empty-string fallback on `VITE_YVP_APP_KEY`, passing `undefined` from an unset env var as `''` at the prop boundary so the provider guard triggers correctly. |
| .changeset/missing-app-key-message.md | Correctly marks both `@youversion/platform-react-hooks` and `@youversion/platform-react-ui` as minor bumps; description is accurate. |

<sub>Reviews (10): Last reviewed commit: ["refactor(ui): log missing-appKey guidanc..."](https://github.com/youversion/platform-sdk-react/commit/434e9fe11cc725f130f2d5f1f47c27836426c959) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37035699)</sub>


packages/ui/src/i18n/locales/es.json
cameronapak, line 1: Hey @camrun91, what do we do about locales that we add that aren't yet Crowdin translated? Do we put in potential strings or leave those in English in these locale files?

camrun91: What do you mean new locals? If you are talking adding new strings the process will be to add them to the file in the localization repo. This is a good test to find out the best way to do this as the english source files shouldn't need to be translated and we need a process to make this quicker

cameronapak: Okay, so does the localization repo need a PR and that PR merged in before I can work on this? @camrun91

cameronapak:
@Dustin-Kelley can you review this? I'm no longer adding new strings to this and re-using old ones

P.S. I'll learn the new strings process and share that with you soon

packages/ui/src/i18n/locales/es.json
camrun91: I think this got handled correct?

packages/hooks/src/context/YouVersionProvider.tsx
```diff
@@ -79,6 +79,18 @@ export function YouVersionProvider(
...
+  // package's provider catches this earlier and renders a styled message
+  // instead; this throw is the baseline guarantee for hooks-only consumers.
+  if (!appKey?.trim()) {
+    throw new Error(
+      'YouVersionProvider: a non-empty "appKey" is required. If you load it from an ' +
+        'environment variable, make sure it is set and restart your dev server.',
+    );
+  }
+
   const resolvedTheme = useResolvedTheme(theme);
```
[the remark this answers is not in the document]
cameronapak: Addressed in 2ddd583. Split the provider: the missing-`appKey` throw now lives in a thin `YouVersionProvider` wrapper, and all hooks moved into an inner `YouVersionProviderInner` component. The hook-bearing component is never entered with an invalid key, so hook order stays consistent across renders. Lint, typecheck, and all 278 hooks tests pass.

examples/vite-react/src/ThemedApp.tsx
```diff
@@ -4,15 +4,17 @@ import App from './App';
...
+  const appKey = import.meta.env.VITE_YVP_APP_KEY;
   const apiHost = import.meta.env.VITE_YVP_API_HOST ?? 'api.youversion.com';
   const authRedirectUrl = import.meta.env.VITE_YVP_AUTH_REDIRECT_URL ?? window.location.origin;
 
   return (
     <YouVersionProvider
       theme={theme}
       apiHost={apiHost}
-      appKey={appKey}
+      appKey={appKey ?? ''}
```
Dustin-Kelley, line 17: nitpick(non-blocking): I don't know too much about this codebase yet so I could be totally wrong. If it's possible I usually like to guard against the empty `appKey` before it gets passed to this provider so you can always expect it to be there within the provider. Not sure if that's possible in this case though. Either way it gets the job done!

Dustin-Kelley (dismissed):

Dustin-Kelley (approved):
💯

PR #283, opened 2026-07-10
Wire BibleReader highlights to the highlights API (YPE-1034) (1/4)

cameronapak:
## Summary

BibleReader renders a signed-in user's server highlights, dark-launched behind the `HIGHLIGHTS_LIVE` flag (off by default). Restored from consolidated #288 and rebased onto current main.

Stack: **1/4 (this PR)** → 2/4 #285 → 3/4 #286 → 4/4 #294 · review the stack bottom-up

## Changes

1. BibleReader shows the user's highlights from the highlights API when `HIGHLIGHTS_LIVE` is enabled — no visible change while the flag is off.
2. Highlight data refetches when the flag flips on, clears when it flips off, and latest-wins on overlapping responses (`useApiData` hardening in hooks).
3. USFM range utilities map highlight verse ranges for rendering.
4. ADR + glossary document the server-only highlights decision (no local persistence).

## Test plan

- `pnpm build && pnpm typecheck && pnpm lint && pnpm test` at this branch: all green (core 315 / hooks 281 / ui 211).
- Storybook integration tests run in CI (not runnable in the local sandbox).

<h3>Greptile Summary</h3>

This PR wires BibleReader's highlight system to the server-side highlights API behind an internal `HIGHLIGHTS_LIVE` dark-launch flag (currently `false`), replacing the temporary localStorage store with a server-only model. It also hardens `useApiData` with a monotonic sequence number to suppress stale responses from both effect-initiated and refetch-initiated requests, and exports `YouVersionAuthContext` for consumers that must tolerate a missing auth provider.

- **`useApiData` hardening**: replaces per-effect `canceled` closures with a shared `requestSeqRef`; adds `enabled` to effect deps so a `false\u2192true` flip fires the fetch; clears `data`/`error` on disable to prevent stale account data from lingering across sessions.
- **`useBibleReaderHighlights`**: new seam hook with optimistic overlay, scope-gated confirmed-verse drain, auth-tolerant context read, and dark-launch gate. All previous-review bugs are fixed and regression-tested.
- **`usfm-ranges.ts`**: pure utility for collapsing verse lists into USFM range strings (`JHN.3.16-18`), fully unit-tested.

Safe to merge — the entire highlight path is dark-launched behind HIGHLIGHTS_LIVE=false, so no changed behavior reaches production until the flag flips.

The two observations are both narrow and flag-gated: the stale captured-closure refetch interaction only matters while HIGHLIGHTS_LIVE is on, and the module-level flag state is only mutable via an @internal test helper. The core logic is correct and thoroughly tested.

useApiData.ts and use-bible-reader-highlights.ts warrant a second look before HIGHLIGHTS_LIVE is flipped: the stale-closure refetch race should be resolved in PR 2 as planned.

| Filename | Overview |
|----------|----------|
| packages/hooks/src/useApiData.ts | Replaced per-effect `canceled` closure with a monotonic `requestSeqRef` sequence number; adds `enabled` to effect deps; clears data/error on disable. Subtle interaction with stale captured closures from `createHighlight` flagged as a known deferred concurrency issue. |
| packages/ui/src/components/use-bible-reader-highlights.ts | New hook: optimistic overlay with snapshot/revert, scope-gated confirmed-verse drain, auth-tolerant context read, dark-launch gate. All previous-review findings addressed. |
| packages/ui/src/lib/feature-flags.ts | New dark-launch module for HIGHLIGHTS_LIVE=false. Module-level mutable state is safe today (only used by 'use client' modules) but would be problematic if imported server-side. |
| packages/ui/src/components/bible-reader.tsx | Removes localStorage highlight store, delegates to useBibleReaderHighlights. Mechanical and clean. |
| packages/hooks/src/context/index.ts | Exports YouVersionAuthContext via context/index.ts per the established pattern. |
| packages/ui/src/lib/usfm-ranges.ts | New utility collapsing verse lists into contiguous USFM range strings. Well-tested. |
| packages/hooks/src/useApiData.test.tsx | New tests covering enabled false→true transitions and stale response suppression with deferred promises. |
| packages/ui/src/components/use-bible-reader-highlights.test.tsx | Comprehensive unit tests including regression coverage for all three previously-reviewed bugs. |
| packages/ui/src/components/use-bible-reader-highlights.integration.test.tsx | Integration tests against the real useHighlights/useApiData stack covering auth-resolves-after-mount and sign-out. |

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant BR as BibleReader
    participant UBH as useBibleReaderHighlights
    participant UH as useHighlights
    participant UAD as useApiData
    participant API as Highlights API
    Note over BR,API: Flag ON + signed-in
    BR->>UBH: "mount {versionId, book, chapter}"
    UBH->>UH: useHighlights passage_id JHN.3 enabled true
    UH->>UAD: fetchFn getHighlights
    UAD->>API: GET highlights JHN.3
    API-->>UAD: verse highlights data
    UAD-->>UBH: highlightedVerses
    UBH-->>BR: highlightedVerses map
    Note over BR,API: User taps color
    BR->>UBH: apply color verses
    UBH->>UBH: patchOverlay optimistic
    UBH->>UH: createHighlight range USFM
    UH->>API: POST highlights
    API-->>UH: 201 Created
    UH->>UAD: refetch
    UAD->>API: GET highlights JHN.3
    API-->>UAD: updated data
    UBH->>UBH: drain confirmed overlay entries
    Note over BR,API: Sign-out
    UBH->>UH: enabled false
    UAD->>UAD: setData null
    UBH-->>BR: highlightedVerses empty
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
sequenceDiagram
    participant BR as BibleReader
    participant UBH as useBibleReaderHighlights
    participant UH as useHighlights
    participant UAD as useApiData
    participant API as Highlights API
    Note over BR,API: Flag ON + signed-in
    BR->>UBH: "mount {versionId, book, chapter}"
    UBH->>UH: useHighlights passage_id JHN.3 enabled true
    UH->>UAD: fetchFn getHighlights
    UAD->>API: GET highlights JHN.3
    API-->>UAD: verse highlights data
    UAD-->>UBH: highlightedVerses
    UBH-->>BR: highlightedVerses map
    Note over BR,API: User taps color
    BR->>UBH: apply color verses
    UBH->>UBH: patchOverlay optimistic
    UBH->>UH: createHighlight range USFM
    UH->>API: POST highlights
    API-->>UH: 201 Created
    UH->>UAD: refetch
    UAD->>API: GET highlights JHN.3
    API-->>UAD: updated data
    UBH->>UBH: drain confirmed overlay entries
    Note over BR,API: Sign-out
    UBH->>UH: enabled false
    UAD->>UAD: setData null
    UBH-->>BR: highlightedVerses empty
```

</a>

<sub>Reviews (4): Last reviewed commit: ["refactor: apply non-blocking review sugg..."](https://github.com/youversion/platform-sdk-react/commit/a36263a6ed49d5e0cf48af98834172cbbda6b75a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=44159815)</sub>

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -1,0 +1,85 @@
+# Ubiquitous Language
+
+Glossary of domain terms for the YouVersion Platform SDK. Terms here are
+canonical: code, docs, and conversation should use them exactly.
+
+## Highlight
+
+A user-owned color marking on a Bible passage, stored on the user's
+YouVersion account. Highlights are **account data, not device data**: they
+require an authenticated user and are never persisted locally by the SDK
+(decided in YPE-1034, superseding the temporary localStorage store from
+YPE-642 / ADR-001).
+
+Identified by the pair (**Bible version**, **passage**). A highlight has
+exactly one **color**.
+
+## Passage
+
+A verse or contiguous verse range in one chapter of one Bible version,
+identified by a USFM string (`JHN.3.16`, `JHN.3.16-18`). A chapter USFM
+(`JHN.3`) is a passage *scope* used for querying, not a highlightable unit.
+
+## Bible version
+
+A translation/edition of the Bible, identified by a numeric id. The SDK
+calls this `version_id`; the highlights wire API calls the same value
+`bible_id`. The SDK name is canonical in public types; mapping happens at
+the API boundary only.
+
+## Color
+
+A highlight's fill, a 6-character lowercase hex string without `#`
+(`fff9b1`). Uppercase input is accepted and normalized at the API boundary.
+
+## Verse selection
+
+The ephemeral set of verses the reader has tapped in `BibleReader`. Drives
+the verse action popover. Never persisted; cleared on navigation.
+
+## Self-contained mode
+
+The default `BibleReader` posture (YPE-1034): the reader fetches and writes
+highlights itself through the SDK's own auth session. The color row is
+always offered; tapping a color when the user lacks a session or the
+highlights permission enters the highlight auth flow (July 9 2026 sync,
+superseding the earlier hide-when-signed-out idea).
+
+## Highlights permission
+
+The per-app grant that authorizes highlight reads/writes for a user. It is
+**not an OIDC scope**: it travels as `requested_permissions[]=highlights`
+alongside `scope` at authorize time (PR #280) and is granted via a data
+exchange consent. Permissions are open-ended strings, never enums (more
+arrive later, e.g. verse notes). The **server is the source of truth** for
+whether it is granted; any client-side permission cache is optimistic only.
+
+## Pending highlight
+
+The user's stashed highlight intent (verses + color + timestamp) while the
+highlight auth flow is in flight. Survives a redirect round-trip via
+sessionStorage; expires stale (~10 min) so an abandoned round-trip can
+never apply a highlight during a later sign-in. Discarded on decline,
+cancel, or failure.
+
+## Highlight auth flow
+
+The state machine (see the React Web auth state machine doc) that turns a
+color tap into an applied highlight across two user paths: one-fell-swoop
+(sign-in requesting the highlights permission together) and just-in-time
+(already signed in, permission confirm dialog → data exchange grant).
+Cancellation at any point keeps the verse selection intact and discards
+only the pending highlight.
+
+## Controlled mode
+
+A planned `BibleReader` posture (YPE-3705): the host application supplies
+rendered highlights as data and receives highlight-intent events, and the
+reader makes no highlight network calls. The host owns auth, persistence,
+and conflict rules. Used by native hosts (RN Expo SDK) embedding the web
+reader.
+
+## Verse action popover
+
+The floating action bar (YPE-642) that appears over a verse selection,
+offering highlight colors, copy, and share.
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -39,45 +39,11 @@ the verse action popover. Never persisted; cleared on navigation.
 
 ## Self-contained mode
 
-The default `BibleReader` posture (YPE-1034): the reader fetches and writes
-highlights itself through the SDK's own auth session. The color row is
-always offered; tapping a color when the user lacks a session or the
-highlights permission enters the highlight auth flow (July 9 2026 sync,
-superseding the earlier hide-when-signed-out idea).
-
-## Highlights permission
-
-The per-app grant that authorizes highlight reads/writes for a user. It is
-**not an OIDC scope**: it travels as `requested_permissions[]=highlights`
-alongside `scope` at authorize time (PR #280) and is granted via a data
-exchange consent. Permissions are open-ended strings, never enums (more
-arrive later, e.g. verse notes). The **server is the source of truth** for
-whether it is granted; any client-side permission cache is optimistic only.
-
-## Pending highlight
-
-The user's stashed highlight intent (verses + color + timestamp) while the
-highlight auth flow is in flight. Survives a redirect round-trip via
-sessionStorage; expires stale (~10 min) so an abandoned round-trip can
-never apply a highlight during a later sign-in. Discarded on decline,
-cancel, or failure.
-
-## Highlight auth flow
-
-The state machine (see the React Web auth state machine doc) that turns a
-color tap into an applied highlight across two user paths: one-fell-swoop
-(sign-in requesting the highlights permission together) and just-in-time
-(already signed in, permission confirm dialog → data exchange grant).
-Cancellation at any point keeps the verse selection intact and discards
-only the pending highlight.
-
-## Controlled mode
-
-A planned `BibleReader` posture (YPE-3705): the host application supplies
-rendered highlights as data and receives highlight-intent events, and the
-reader makes no highlight network calls. The host owns auth, persistence,
-and conflict rules. Used by native hosts (RN Expo SDK) embedding the web
-reader.
+The `BibleReader` posture (YPE-1034): the reader fetches and writes
+highlights itself through the SDK's own auth session. Highlight behavior is
+gated on the internal `HIGHLIGHTS_LIVE` dark-launch flag and an
+authenticated user — while the flag is off or the user has no session, the
+reader is inert: no fetches, no writes, nothing rendered from the API.
 
 ## Verse action popover
 
--- a/packages/hooks/src/useApiData.test.tsx
+++ b/packages/hooks/src/useApiData.test.tsx
@@ -5,22 +5,6 @@ import { act, renderHook, waitFor } from '@testing-library/react';
 import { describe, expect, it, vi } from 'vitest';
 import { useApiData, type UseApiDataOptions } from './useApiData';
 
-type Deferred<T> = {
-  promise: Promise<T>;
-  resolve: (value: T) => void;
-  reject: (reason: unknown) => void;
-};
-
-function createDeferred<T>(): Deferred<T> {
-  let resolve!: (value: T) => void;
-  let reject!: (reason: unknown) => void;
-  const promise = new Promise<T>((res, rej) => {
-    resolve = res;
-    reject = rej;
-  });
-  return { promise, resolve, reject };
-}
-
 describe('useApiData — enabled transitions', () => {
   it('fetches when enabled flips from false to true with unchanged deps', async () => {
     const fetchFn = vi.fn().mockResolvedValue('payload');
@@ -68,9 +52,9 @@ describe('useApiData — enabled transitions', () => {
 
 describe('useApiData — stale responses (latest wins)', () => {
   it('ignores a stale refetch response that resolves after a newer fetch', async () => {
-    const deferreds: Deferred<string>[] = [];
+    const deferreds: PromiseWithResolvers<string>[] = [];
     const fetchFn = vi.fn(() => {
-      const deferred = createDeferred<string>();
+      const deferred = Promise.withResolvers<string>();
       deferreds.push(deferred);
       return deferred.promise;
     });
@@ -112,9 +96,9 @@ describe('useApiData — stale responses (latest wins)', () => {
   });
 
   it('ignores a stale error from an invalidated request', async () => {
-    const deferreds: Deferred<string>[] = [];
+    const deferreds: PromiseWithResolvers<string>[] = [];
     const fetchFn = vi.fn(() => {
-      const deferred = createDeferred<string>();
+      const deferred = Promise.withResolvers<string>();
       deferreds.push(deferred);
       return deferred.promise;
     });
--- a/packages/hooks/src/useApiData.test.tsx
+++ b/packages/hooks/src/useApiData.test.tsx
@@ -1,0 +1,175 @@
+/**
+ * @vitest-environment jsdom
+ */
+import { act, renderHook, waitFor } from '@testing-library/react';
+import { describe, expect, it, vi } from 'vitest';
+import { useApiData, type UseApiDataOptions } from './useApiData';
+
+type Deferred<T> = {
+  promise: Promise<T>;
+  resolve: (value: T) => void;
+  reject: (reason: unknown) => void;
+};
+
+function createDeferred<T>(): Deferred<T> {
+  let resolve!: (value: T) => void;
+  let reject!: (reason: unknown) => void;
+  const promise = new Promise<T>((res, rej) => {
+    resolve = res;
+    reject = rej;
+  });
+  return { promise, resolve, reject };
+}
+
+describe('useApiData — enabled transitions', () => {
+  it('fetches when enabled flips from false to true with unchanged deps', async () => {
+    const fetchFn = vi.fn().mockResolvedValue('payload');
+
+    const { result, rerender } = renderHook(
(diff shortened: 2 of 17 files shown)


packages/ui/src/components/use-bible-reader-highlights.ts
```diff
@@ -0,0 +1,278 @@
...
+      const next = { ...current };
+      for (const verse of confirmed) {
+        if (verse in next) {
+          delete next[verse];
+          changed = true;
+        }
+      }
+      return changed ? next : current;
+    });
+  }, [highlights]);
```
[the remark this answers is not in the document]
cameronapak: Fixed in 65f61b2: writes now capture the overlay scope at start and only enroll verses into the confirmed set if the scope still matches at settle time; the set is also cleared on scope change alongside the overlay reset. Added a regression test covering the JHN.3.16 → JHN.4.16 scenario (confirmed red before the fix).

packages/ui/src/components/use-bible-reader-highlights.ts
```diff
@@ -0,0 +1,304 @@
...
+          console.error(
+            `[YouVersion SDK] Failed to apply highlight (version ${versionId}, ` +
+              `passage ${failure.error.passageId}, color ${normalizedColor})`,
+            failure.error,
+          );
+        }
+        // Partial failures revert the whole optimistic batch; the refetch that
+        // any successful create triggered repaints the server's truth.
+        revertOverlay(verses, snapshot);
+      })();
```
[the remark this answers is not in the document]
cameronapak: Fixed in c636c50: both failure-path reverts are now gated on the same scopeAtWrite check as the success path (nothing to revert after navigation — the scope reset already wiped that overlay). Error logging stays unconditional. Added a regression test with a failing stale write; confirmed red before the fix.

jaredhightower-youversion (commented):
Did a review using /ponytail-review skill and it found a few things:

jaredhightower-youversion (commented):
Did a simplification-focused pass over this PR — not looking for bugs (the earlier greptile threads covered those and they're all resolved; I haven't re-raised any of them), just places where the diff could be smaller and easier to maintain. Found five. None of them are blockers, but together they'd cut roughly 70 lines and one dependency, and each comment below explains what's going on and includes the replacement code so it should be quick to action.

jaredhightower-youversion (commented):
Added [conventional comments](https://conventionalcomments.org) format, all non-blocking.

packages/ui/package.json
```diff
@@ -53,12 +53,13 @@
     "@radix-ui/react-use-controllable-state": "^1.2.2",
     "@youversion/platform-core": "workspace:*",
     "@youversion/platform-react-hooks": "workspace:*",
+    "better-result": "2.9.2",
```
jaredhightower-youversion, line 56: **suggestion (non-blocking):** Replace the `better-result` dependency with native promise handling.

Tracing where it's actually used, it comes down to one pattern: `Result.tryPromise` inside the two write callbacks in `use-bible-reader-highlights.ts` (`apply` and `remove`). What that code needs is "run all the requests, collect which ones failed and why" — and native promises express that directly by mapping each rejection to a `{ passageId, cause }` record:

```ts
const failures = (
  await Promise.all(
    buildPassageIds(book, chapter, verses).map((passageId) =>
      createHighlight({ version_id: versionId, passage_id: passageId, color: normalizedColor })
        .then(() => null, (cause) => ({ passageId, cause })),
    ),
  )
).filter((f) => f !== null);
```

Same failure list, same partial-failure semantics (successes still trigger their refetch, failures still revert the overlay), no new package in consumers' install graphs. For an SDK other teams install, every production dependency is supply-chain and version-conflict surface, so I'd hold the bar at "only add one when the platform can't do it" — and here the platform can. This also makes `BibleReaderHighlightError` unnecessary (separate comment).

packages/ui/src/components/use-bible-reader-highlights.ts
```diff
@@ -0,0 +1,309 @@
...
+  readonly cause: unknown;
+
+  constructor(operation: 'apply' | 'remove', passageId: string, cause: unknown) {
+    super(`Highlight ${operation} failed for ${passageId}`);
+    this.name = 'BibleReaderHighlightError';
+    this.operation = operation;
+    this.passageId = passageId;
+    this.cause = cause;
+  }
+}
```
jaredhightower-youversion, line 47: **suggestion (non-blocking):** Drop the `BibleReaderHighlightError` class in favor of a plain `{ passageId, cause }` object.

Following its lifecycle, nothing ever treats it as an error *type*: it isn't caught by type anywhere, isn't rethrown, isn't exported, and never crosses the hook's boundary. Its only job is to ferry `passageId` and `cause` from the failed request to the `console.error` a few lines later — a plain object does that with zero ceremony:

```ts
(cause) => ({ passageId, cause })
```

It also drags subtle weight: a manual `cause` field that shadows ES2022's native `Error.cause`, and a message string nothing reads. If `better-result` goes (per my package.json comment), this class has no reason to exist. When a later PR genuinely needs a typed error — say PR 2's 401/403 handling wants `instanceof` checks — that's the right moment to introduce one; right now it's structure without a consumer.

packages/hooks/src/useApiData.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+function createDeferred<T>(): Deferred<T> {
+  let resolve!: (value: T) => void;
+  let reject!: (reason: unknown) => void;
+  const promise = new Promise<T>((res, rej) => {
+    resolve = res;
+    reject = rej;
+  });
+  return { promise, resolve, reject };
+}
```
jaredhightower-youversion: **nitpick (non-blocking):** The hand-rolled `Deferred<T>` + `createDeferred` helper is now built into the language.

ES2024 shipped `Promise.withResolvers()`, available on this repo's Node floor (≥ 22.13):

```ts
const { promise, resolve, reject } = Promise.withResolvers<string>();
```

It returns the exact shape `createDeferred` builds — same field names — so each `deferreds.push(...)` site works unchanged and the ~15 lines of helper at the top of the file can go. Less scaffolding for the next reader to verify before getting to the actual test logic.

**praise:** The stale-response sequencing coverage in this file is genuinely good — the deferred-resolution ordering pins the latest-wins behavior precisely.

packages/ui/src/lib/feature-flags.ts
```diff
@@ -0,0 +1,31 @@
...
+ * SDK-internal switches, not consumer API. Flip the default here when a
+ * feature is ready to launch.
+ */
+
+/**
+ * Dark-launch default for server-backed highlights in BibleReader (YPE-1034).
+ * While `false`, the reader's highlight color row is inert: no highlight
+ * fetches, no writes, nothing rendered from the API.
+ */
+export const HIGHLIGHTS_LIVE = false;
```
jaredhightower-youversion, line 6: **nitpick (non-blocking):** 31 lines for one boolean — same API fits in ~8.

The const + mutable let + getter + setter shape reads like the start of a feature-flag framework, but for an internal dark-launch switch that ships hardcoded and only gets flipped by tests/Storybook:

```ts
/**
 * Dark-launch default for server-backed highlights in BibleReader (YPE-1034).
 * While `false`, the reader's highlight color row is inert: no fetches, no
 * writes, nothing rendered from the API. Flip here to launch.
 */
export const HIGHLIGHTS_LIVE = false;

let highlightsLive = HIGHLIGHTS_LIVE;

export const isHighlightsLive = (): boolean => highlightsLive;

/** @internal Test/Storybook only — launch by changing HIGHLIGHTS_LIVE. */
export const setHighlightsLive = (value: boolean): void => void (highlightsLive = value);
```

No caller changes — purely collapsing doc-comment and function ceremony. Worth doing because this file is the template the next flag will copy.

CONTEXT.md
```diff
@@ -0,0 +1,85 @@
+# Ubiquitous Language
```
jaredhightower-youversion, line 1: **thought (non-blocking):** About half this glossary documents behavior that doesn't exist yet in this PR.

**Pending highlight** (sessionStorage stash + ~10-min expiry), **Highlight auth flow** (the state machine), **Highlights permission** (the `requested_permissions` grant), and **Controlled mode** (YPE-3705) all describe PR-2-and-later work with no code behind them in this diff. The risk of documenting ahead of implementation is silent drift: PR 2 will make real decisions (maybe the expiry isn't 10 minutes, maybe the flow shape changes), and this file won't fail a test or build when it stops being true — it just quietly becomes wrong. A canonical-terms doc that's wrong is worse than no doc.

Two ways to resolve, either works: trim to the terms this PR gives meaning to (Highlight, Passage, Bible version, Color, Verse selection, Verse action popover) and let each future PR add entries as behavior lands — or split the full glossary into its own docs PR so it's reviewed as a design artifact rather than riding along with implementation. My preference is the first: glossary entries born in the same PR as their code can't drift.

packages/hooks/src/useApiData.ts
```diff
@@ -24,48 +24,62 @@ export function useApiData<T>(
...
   }, [fetchFn, enabled]);
 
   const refetch = useCallback(() => {
     fetchData();
   }, [fetchData]);
 
+  // `enabled` rides alongside the caller-supplied deps so a false→true flip
+  // (e.g. auth resolving after mount, with the caller's deps unchanged)
+  // actually triggers the fetch.
   useEffect(() => {
```
jaredhightower-youversion, line 75: [question](https://conventionalcomments.org/): At some point in the future, are we thinking about converting a lot of these use effects to use TanStack queries?

Dustin-Kelley: ohh I haven't heard of BrowserStack queries! Is that something you guys are using on other platforms? We had talked about pulling in Tanstack Query at some point.

packages/ui/package.json
Dustin-Kelley: @cameronapak are you using `better-result` in upcoming PR's? If so then it's worth the keep.

packages/ui/src/components/use-bible-reader-highlights.ts
Dustin-Kelley, line 1: Gosh this hook is so big LOL! This would be a great candidate for TQ in the future. But it's doing it's job 🤙

Dustin-Kelley (commented):
I'm pretty happy with this! I think a lot of lines will be removed when/if we add a proper query library in the future. Just address Jareds comment's and I can approve.

packages/ui/src/components/use-bible-reader-highlights.ts
cameronapak: TQ = TanStack Query

packages/ui/package.json
cameronapak: Checked the stack before deciding: PR 285 uses better-result in new code (Result.tryPromise, Result.isError) and 294 references it as well, so keeping it for now per Dustin's criterion. Jared, your supply-chain bar for an SDK is the right default and worth revisiting once the stack lands and the full usage is visible in one place.

packages/ui/src/components/use-bible-reader-highlights.ts
cameronapak: Keeping for now since PR 285 reshapes these exact write paths. Will revisit dropping it (or introducing a real typed error if 401/403 handling needs one) there rather than churning the stack here.

packages/hooks/src/useApiData.test.tsx
cameronapak: Applied in a36263a: helper deleted, both sites use Promise.withResolvers now. Needed a one-line ES2024 lib bump in the hooks tsconfig since the shared config sits at ES2022. Thanks for the note on the sequencing coverage.

packages/ui/src/lib/feature-flags.ts
cameronapak: Applied in a36263a: collapsed to your shape, same three exports, no caller changes.

CONTEXT.md
cameronapak: Went with your first option in a36263a: removed the four future-work entries (pending highlight, highlight auth flow, highlights permission, controlled mode) and rewrote Self-contained mode to describe only what this PR ships. Each later PR in the stack adds its terms as the behavior lands.

packages/hooks/src/useApiData.ts
cameronapak: Assuming TanStack Query: it's on the radar, and useApiData plus this hook are the obvious first candidates when we take that on. Did not want to couple this stack to a new dependency decision.

packages/ui/src/components/use-bible-reader-highlights.ts
cameronapak: Agreed, TanStack Query would shrink this a lot. Noted for the future refactor.

packages/ui/package.json
cameronapak: Yes, using better result in more places in the following PRs!

Dustin-Kelley (approved):

packages/hooks/src/useApiData.ts
jaredhightower-youversion: 🤣 That was my back on the "BrowserStack" comment; I meant TanStack.

PR #285, opened 2026-07-10
Highlights auth/write flow: xstate statechart + sign-in dialog (2/4)

cameronapak:
## Summary

The full highlights interaction flow — auth, data-exchange permission, writes, resume — is an xstate v5 statechart, with a SignInDialog for signed-out color taps. This presents the machine in its final reviewed form; the staging fixes from old #287 are embedded, so that PR stays closed.

Stack: 1/4 #283 → **2/4 (this PR)** → 3/4 #286 → 4/4 #294 · restored from #288

## Changes

1. Tapping a highlight color while signed out opens SignInDialog (ported from Swift `SignInWithYouVersionView` as a dialog, not a bottom sheet) before OAuth; `appName` and `signInPromptMessage` are new optional statics on `YouVersionPlatformConfiguration`.
2. `bible-reader-highlights-machine.ts` (xstate v5) owns the whole flow — fetch, permission, write batches, pending-write resume, sign-out — and `useBibleReaderHighlights` becomes a thin adapter; mermaid statechart at `docs/highlight-flow-statechart.md`.
3. Highlight writes settle per sub-write and refetch on failure, so one failed batch can't strand optimistic state (absorbed from #287). Permission-lost writes stash their intents as a LIST (verse-level last-wins), so multiple failed batches all survive the re-grant.
4. Deleted highlights can't resurrect from stale replica reads — remove-overlay entries persist until scope change, sign-out, or a newer write.
5. Granted-permissions cache is user-scoped `{userId, permissions}` in localStorage.
6. Active-swatch remove affordance temporarily renders an X; the checkmark lands in 3/4 (one code comment references it early — intentional).
7. Out of scope by design (land in 4/4 #294): initiating-user grant binding and `highlightsInteractive` gating. (`xstate` is exact-pinned to 5.32.5 in this PR.)

**Start here:** change 2 — the statechart is the review unit; the mermaid doc mirrors its states.

## Test plan

- `pnpm build && pnpm typecheck && pnpm lint && pnpm test` at this branch: all green (889 tests — core 342 / hooks 284 / ui 263).
- `pnpm check:i18n` passes (warnings only: es/fr keys pending upstream sync).
- locale-ownership CI: **green**. All 15 i18n keys this PR uses landed on main via the platform-localization syncs (#298, #299); the PR no longer touches any locale file. es/fr translations arrive via Crowdin sync later (check:i18n warn-only).

<h3>Greptile Summary</h3>

This PR wires the full highlights auth flow — signed-out color tap → sign-in dialog → OAuth redirect → pending resume; signed-in without permission → just-in-time data-exchange flow — into an xstate v5 statechart (`bible-reader-highlights-machine.ts`), replacing the hand-rolled effect orchestration. `useBibleReaderHighlights` becomes a thin adapter; `SignInDialog` is ported from the Swift SDK; and fixes absorbed from staging (#287, #288) are embedded here.

- **Statechart (`bible-reader-highlights-machine.ts`)**: parallel `flow` + `writer` regions handle the auth dialogs, serialized write queue, per-verse ownership tokens, and the vapor fix (remove overlay entries never retire, preventing stale-replica ghosts).
- **Pending highlight stash (`pending-highlight.ts`)**: upgraded from single-slot to a list with verse-level last-wins merge so multiple concurrent 401-failing writes each survive the re-grant.
- **Permission cache (`YouVersionPlatformConfiguration`)**: user-scoped `{userId, permissions}` in localStorage; seeded from both sign-in and data-exchange callbacks; invalidated on 401/403; cleared on sign-out.
- **`useHighlights` refetch removal**: `createHighlight`/`deleteHighlight` no longer auto-refetch; callers coalesce and call `refetch()` once after the batch settles.

Safe to merge. The statechart, permission cache, and pending-highlight list are all correctly implemented; previous review feedback is fully incorporated.

All major edge cases are handled and tested: concurrent 401-failing writes both survive the re-grant (verse-level last-wins merge), stale-replica ghost highlights are suppressed (remove overlay entries never retire), scope changes synchronously clear ownership tokens so stale writes cannot pollute new-scope state, and the permission cache is correctly user-scoped. The three previously flagged bugs have verified fixes with new direct-machine tests. Deferred items are explicitly documented with follow-up tracking.

No files require special attention. bible-reader-highlights-machine.ts is the review unit per the PR description and holds up well under scrutiny.

| Filename | Overview |
|----------|----------|
| packages/ui/src/components/bible-reader-highlights-machine.ts | New xstate v5 statechart: parallel flow/writer regions, per-verse ownership tokens, vapor fix, pending-resume logic. Logic is correct and well-tested; deferred items (token-expiry 401 misrouting, resume-write failure handling) are clearly documented. |
| packages/ui/src/lib/pending-highlight.ts | Upgraded from single-slot to a list with per-entry TTL and verse-level last-wins merge. appendPendingHighlight correctly handles concurrent 401-failing writes; expiry and stale-cleanup logic are sound. |
| packages/core/src/YouVersionPlatformConfiguration.ts | New user-scoped granted-permissions cache ({userId, permissions} in localStorage), appName/signInPromptMessage statics. Cache is correctly scoped, cleared on sign-out, and guarded against user-mismatch reads. |
| packages/ui/src/components/use-bible-reader-highlights.ts | Reduced to a thin adapter: spawns the machine once, keeps a latest-ref bag for services, feeds AUTH_CHANGED/SCOPE_CHANGED/HIGHLIGHTS_UPDATED events, and derives dialog state via useSelector. Clean and correct. |
| packages/core/src/data-exchange.ts | New data-exchange client: mints short-lived token, builds consent URL, parses callback params, handles granted return (saves permissions, strips URL). Logic mirrors the Swift SDK contract correctly. |
| packages/hooks/src/useHighlightAuthActions.ts | New hook exposing hasHighlightsPermission, invalidateHighlightsPermission, consumeDataExchangeReturn, startSignInForHighlights, startDataExchangeForHighlights. Reads context defensively so it never throws when no auth provider is mounted. |
| packages/hooks/src/useHighlights.ts | Removed auto-refetch from createHighlight/deleteHighlight; callers now coalesce and call refetch() once after a write batch settles. Breaking change documented in JSDoc; changeset acknowledges it. |
| packages/ui/src/components/sign-in-dialog.tsx | New SignInDialog component ported from Swift SDK. Presentational only, all i18n copy routed through t(), promptMessage hidden when unset. Callbacks delegated to parent. |
| packages/ui/src/components/bible-reader-highlights-machine.test.ts | New direct machine tests covering writeIntent lifecycle, scope-change clearing, and the multi-401 batch scenario (both intents survive and re-apply). |
| packages/core/src/Users.ts | Sign-in callback now seeds the granted-permissions cache from granted_permissions URL params. saveUserInfo is correctly called before saveGrantedPermissions so the user-scope guard is satisfied at seed time. |

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant U as User
    participant BR as BibleReader
    participant M as HighlightsMachine
    participant SS as SessionStorage
    participant LS as LocalStorage
    participant API as YouVersion API

    Note over M: boot to enabled (flow + writer in parallel)

    U->>BR: tap color (signed out)
    BR->>M: TAP_COLOR
    M->>SS: stashPendingHighlight
    M-->>BR: "lastTapOutcome = flow"
    Note over M: flow to signInDialog

    U->>BR: confirm sign-in
    BR->>M: CONFIRM_SIGN_IN
    M->>API: startSignInForHighlights (full-page redirect)

    API-->>BR: redirect back with code + granted_permissions
    BR->>LS: saveUserInfo then saveGrantedPermissions
    Note over M: new page load resuming

    M->>SS: consumeDataExchangeReturn (null)
    Note over M: pending + authed + hasPermission to applyPendingHighlight
    M->>SS: readPendingHighlights + clearPendingHighlight
    M->>API: createHighlight
    API-->>M: 200 OK
    M->>API: refetch
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
sequenceDiagram
    participant U as User
    participant BR as BibleReader
    participant M as HighlightsMachine
    participant SS as SessionStorage
    participant LS as LocalStorage
    participant API as YouVersion API

    Note over M: boot to enabled (flow + writer in parallel)

    U->>BR: tap color (signed out)
    BR->>M: TAP_COLOR
    M->>SS: stashPendingHighlight
    M-->>BR: "lastTapOutcome = flow"
    Note over M: flow to signInDialog

    U->>BR: confirm sign-in
    BR->>M: CONFIRM_SIGN_IN
    M->>API: startSignInForHighlights (full-page redirect)

    API-->>BR: redirect back with code + granted_permissions
    BR->>LS: saveUserInfo then saveGrantedPermissions
    Note over M: new page load resuming

    M->>SS: consumeDataExchangeReturn (null)
    Note over M: pending + authed + hasPermission to applyPendingHighlight
    M->>SS: readPendingHighlights + clearPendingHighlight
    M->>API: createHighlight
    API-->>M: 200 OK
    M->>API: refetch
```

</a>

<sub>Reviews (8): Last reviewed commit: ["docs(hooks): contract-note the no-auto-r..."](https://github.com/youversion/platform-sdk-react/commit/353d526b88a0931f484c9e65f6da1ef8c50d0fdf) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=44159803)</sub>


cameronapak:
Superseded by #288, which consolidates the full YPE-1034 stack into a single PR rebased onto main.

packages/ui/src/components/bible-reader.tsx
```diff
@@ -520,8 +524,23 @@ function Content() {
...
+    permissionDialogOpen,
+    onPermissionDialogOpenChange,
+    confirmPermissionDialog,
+    cancelPermissionDialog,
+    signInDialogOpen,
+    confirmSignInDialog,
+    cancelSignInDialog,
   } = useBibleReaderHighlights({ versionId, book, chapter });
 
+  // The color row / clear-highlight affordances only render when the highlights
```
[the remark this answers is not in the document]
cameronapak: Fixed in 75bdf08: the fallback now routes through i18n as `signInAppNameFallback` (added to en/es/fr alongside the other keys this PR temporarily carries; it rides the same platform-localization upstreaming as the rest).

packages/ui/src/components/bible-reader-highlights-machine.ts
```diff
@@ -0,0 +1,780 @@
...
+     */
+    settleWrite: enqueueActions(({ enqueue, event }) => {
+      // Wired only to the processWrite `onDone`; the done event carries `output`
+      // but is not part of the public event union, so read it via a cast.
+      const { op, failures, failedVerses, succeededVerses } = (
+        event as unknown as { output: WriteResult }
+      ).output;
+
+      enqueue.assign(({ context: current }) => {
+        const overlay = { ...current.overlay };
```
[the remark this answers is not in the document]
cameronapak: Confirmed and fixed in 75bdf08. `settleWrite` now releases a verse's `writeIntent` claim when the settling op still owns it (token guard keeps a newer op's claim intact), and `resetForScopeChange` clears `writeIntent` so a stale-scope write settling after navigation can't pollute the new scope's reconcile map under a colliding verse number. Covered by three new direct-machine tests in `bible-reader-highlights-machine.test.ts`.

packages/core/src/SignInWithYouVersionResult.ts
```diff
@@ -23,6 +24,12 @@ export class SignInWithYouVersionResult {
   public readonly name: string | undefined;
   public readonly profilePicture: string | undefined;
   public readonly email: string | undefined;
+  /**
+   * Data-exchange permissions the server reported as granted for this sign-in
+   * (parsed from `granted_permissions` on the callback). Empty when the callback
```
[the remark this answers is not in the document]
cameronapak: Fixed in 75bdf08: `permissions` is now `readonly`. The one post-construction mutation site (`Users.extractSignInResult` callback path) now parses `granted_permissions` before construction and passes it through the constructor.

packages/ui/src/i18n/locales/en.json
```diff
@@ -67,5 +67,19 @@
...
+  "signInParagraph": "{{appName}} wants to connect to your YouVersion Bible App account. This will allow them to see and interact with your Bible App activity. Would you like to proceed?",
+  "signInYesButton": "Yes Please",
+  "signInNoButton": "No Thanks",
+  "signOutQuestion": "Are you sure you want to sign out from YouVersion?",
+  "signOutExplanation": "You'll need to sign in again to access your highlights.",
+  "signOutPendingHighlightsQuestion": "Save your highlights?",
+  "signOutPendingHighlightsExplanation": "Some of your highlights haven't been saved yet, and they will be lost if you sign out. Do you want to sign out anyway?",
+  "signOutPendingHighlightsConfirm": "Sign out anyway",
+  "youVersionPlatformLogoAriaLabel": "YouVersion Platform"
 }
```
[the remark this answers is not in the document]
cameronapak: Known and intentional — this PR temporarily carries its new i18n keys (now 15 with `signInAppNameFallback`) and the Locale Ownership check is expected red until youversion/platform-localization#24 merges and the sync-bot PR lands; then this PR drops its locale edits. Called out in the PR description's test plan.

packages/core/src/YouVersionPlatformConfiguration.ts
```diff
@@ -73,6 +75,109 @@ export class YouVersionPlatformConfiguration {
...
+  private static readonly grantedPermissionsKey = 'youversion-platform:granted-permissions';
+
+  /** The id of the user the cache is scoped to, or `null` when signed out. */
+  private static get currentUserId(): string | null {
+    return this.storedUserInfo?.id ?? null;
+  }
+
+  /**
+   * Reads the stored `{ userId, permissions }` entry, or `null` when absent,
+   * malformed, or in the legacy bare-array format (which is treated as absent).
```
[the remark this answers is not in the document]
cameronapak: Fixed in 75bdf08: docstring corrected — `setGrantedPermissions` is a wholesale, de-duplicated replace for the current user (contrasted with `saveGrantedPermissions`'s merge); the false claim about the data-exchange callback is gone. No behavior change; the callback intentionally keeps merge semantics.

packages/ui/src/components/ui/button.tsx
cameronapak, line 1: [question/issue](https://conventionalcomments.org/): Not sure about this. Does this make the rest of the components look wierd/wonky. Or... is primary the same color as background was??

packages/ui/src/components/bible-reader-highlights-machine.ts
```diff
@@ -0,0 +1,797 @@
...
+              }),
+            catch: (cause) =>
+              new BibleReaderHighlightError(
+                'apply',
+                runToPassageId(op.scope.book, op.scope.chapter, run),
+                cause,
+              ),
+          }),
+        ),
+      );
```
cameronapak, line 315: [question](https://conventionalcomments.org/): Does this mean it's trying to highlight one verse at a time versus highlighting a range?

packages/ui/src/lib/pending-highlight.ts
cameronapak, line 1: [question](https://conventionalcomments.org/): worth using zod in here to simplify things? Or does better-result have schema matching?

packages/ui/src/components/ui/button.tsx
cameronapak: Primary was not the same color as background, and that mismatch was the bug: in light mode `--yv-background` is white and `--yv-primary-foreground` is also white, so the old `default` variant (`bg-background` + `text-primary-foreground`) rendered white text on a white surface. This change restores the standard shadcn default (`bg-primary` = gray-50 in light, brand red in dark, with white text).

Blast radius is small: the only components using the `default` variant are the two new dialog CTAs (sign-in confirm, permission confirm), where a filled primary button is the intent. `YouVersionAuthButton` explicitly pins `yv:bg-background` with a code comment for exactly this reason, so the brand button is unaffected. Everything else uses `secondary`/`outline`/`ghost`/`icon` variants. `button.test.tsx` was added in this PR to lock the variant classes down.

packages/ui/src/components/bible-reader-highlights-machine.ts
cameronapak: No, ranges are preserved: `collapseVerseRuns` collapses the selected verses into contiguous runs first, and each run becomes one `createHighlight` call with a range passage id (`runToPassageId` produces e.g. `JHN.3.16-18`). So highlighting verses 16-18 is a single API call. The `Promise.all` fan-out only matters for non-contiguous selections (16, 18, 20 = three runs = three parallel calls). Removes further down are per-verse because deletes target individual passage ids.

packages/ui/src/lib/pending-highlight.ts
cameronapak: Leaning no on both. `better-result` is just a Result/Option library, it has no schema matching. And zod is not currently a dependency of the ui package (it lives in core), so using it here would mean adding a runtime dep to ui for one ~12-line guard on an internal sessionStorage record; reusing core's zod would also cross the ui-imports-core-only-via-public-re-exports boundary for a type that is not API data. The hand-rolled guard stays small even with the multi-entry stash change landing for the thread above, so keeping it as is.

packages/ui/src/components/bible-reader-highlights-machine.ts
```diff
@@ -0,0 +1,797 @@
...
+        // Server says the permission is gone (or never applied): invalidate the
+        // optimistic cache. Server truth wins.
+        enqueue(({ context: current }) =>
+          current.services.current.invalidateHighlightsPermission(),
+        );
+        if (op.kind === 'apply' && op.reprompt) {
+          // Keep this highlight pending and re-prompt the just-in-time dialog.
+          enqueue(() => stashPendingHighlight(opToPending(op)));
+          enqueue.raise({ type: 'PERMISSION_LOST' });
+        }
```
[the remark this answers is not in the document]
cameronapak: Fixed in 7e603b2 by making the stash multi-entry rather than first-wins: a peek guard would only change which tap gets silently dropped, and two pending intents can carry two different colors, which no single-slot merge can represent.

- `pending-highlight.ts` now stores a list. `stashPendingHighlight` still replaces (a fresh tap supersedes everything); new `appendPendingHighlight` merges with verse-level last-wins (new entry's verses are stripped from earlier entries, emptied entries dropped), and expiry is per entry.
- `settleWrite` appends on a permission-denied apply, so sibling 401-failing batches all survive.
- `applyPendingHighlight` resumes every entry first-to-last after the re-grant, each to its own scope, painting only scope-matching entries.

Direct-machine test covers the exact scenario from this comment: two queued applies (color A verses 1-3, color B verses 4-6) both settling 401, asserting both intents survive and both re-apply on resume.

packages/ui/src/components/bible-reader-highlights-machine.ts
```diff
@@ -0,0 +1,805 @@
...
+          enqueue(() => appendPendingHighlight(opToPending(op)));
+          enqueue.raise({ type: 'PERMISSION_LOST' });
+        }
+        // Remove failures invalidate the cache but never re-prompt (the old
+        // deferred wart is fixed here): with no pending highlight, a re-prompt's
+        // post-grant resume was a no-op. The next apply re-enters the flow.
+      } else if (op.kind === 'apply' && op.reprompt) {
+        // Network / 5xx on a user apply: overlay already reverted; drop pending.
+        enqueue(() => clearPendingHighlight());
+      }
```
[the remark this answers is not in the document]
cameronapak: Fixed in cd42fb3 by deleting the clear from the non-permission failure branch. Confirmed the premise: a user apply that reaches the queue via `startApplyWrite` never stashes, so that branch had nothing of its own to drop — it could only wipe a sibling 401 batch's appended intent. All remaining `clearPendingHighlight` call sites are the intended owners (decline/failure data-exchange consume, resume-consume, redirect-launch failure catches, dialog cancel/decline, stale cleanup in the lib). Added the mixed-sequence machine test: write 1 settles 401 (stashes), write 2 settles 5xx — write 1's intent survives and write 2's verses are never stashed.

cameronapak:
thank you sir greppy file-27433cb0f2869f1a96ebc8a71f277c30

packages/core/src/SignInWithYouVersionResult.ts
```diff
@@ -23,6 +24,12 @@ export class SignInWithYouVersionResult {
...
+   * carried none. Additive: existing consumers can ignore it.
+   */
+  public readonly permissions: string[];
```
jaredhightower-youversion: **suggestion (non-blocking):** Drop the unused `permissions` field.

I think we could drop the new `permissions` field on `SignInWithYouVersionResult` entirely (plus the ctor param and the `grantedPermissions` arg threaded into `extractSignInResult`) — nothing ever reads it, and the same data is already persisted via `YouVersionPlatformConfiguration.saveGrantedPermissions()`, which is what all the hooks/UI actually consult. Keeping just the existing cache seeding in `handleAuthCallback` gives the same behavior with less API surface to maintain.

If it's kept intentionally as public SDK surface, note it's currently inconsistent: `Users.refreshTokens()` constructs the result without permissions, so `result.permissions` is always `[]` after a token refresh even for a user who granted permissions.

packages/core/src/YouVersionPlatformConfiguration.ts
```diff
@@ -73,6 +75,110 @@ export class YouVersionPlatformConfiguration {
...
+    this.writeStoredGrants(userId, [...merged]);
+  }
+
+  /**
+   * Replaces the current user's cached grants wholesale with exactly
+   * `permissions` (de-duplicated), discarding whatever was cached before.
+   * Unlike {@link saveGrantedPermissions}, which merges, this is a full
+   * overwrite for callers holding the authoritative set. No-ops when signed out.
+   */
+  public static setGrantedPermissions(permissions: string[]): void {
```
jaredhightower-youversion: **suggestion (non-blocking):** `setGrantedPermissions` has no production callers.

Both `Users.ts` and `data-exchange.ts` use `saveGrantedPermissions`, and this method's only usage is its own unit test. Since the docstring fix in 75bdf08 confirmed the data-exchange callback intentionally merges, could we just delete this method (and its test) and re-add a replace variant if a real caller shows up later?

packages/core/src/data-exchange.ts
```diff
@@ -0,0 +1,148 @@
...
+  constructor(client: ApiClient) {
+    this.client = client;
+  }
+
+  /**
+   * Reads the auth token from the argument or the ambient platform
+   * configuration, mirroring {@link HighlightsClient}. Server-side callers
+   * (no `localStorage`) must pass `lat` explicitly.
+   */
+  private getAuthToken(lat?: string): string {
```
jaredhightower-youversion, line 33: **suggestion (non-blocking):** Share `getAuthToken` with `HighlightsClient` instead of copying it.

I noticed `getAuthToken` here is a verbatim copy of `HighlightsClient.getAuthToken` with only the error message differing — the docstring even says it mirrors it. Could we pull it out into a shared module-level helper like `resolveAuthToken(lat, errorMessage)` that both clients call? That way the localStorage-guard behavior stays in one place if it ever needs to change (and the SSR/no-ambient-token comment from highlights.ts can move to the helper so it isn't lost).

packages/ui/src/components/bible-reader-highlights-machine.ts
```diff
@@ -0,0 +1,808 @@
...
+    super(`Highlight ${operation} failed for ${passageId}`);
+    this.name = 'BibleReaderHighlightError';
+    this.operation = operation;
+    this.passageId = passageId;
+    this.cause = cause;
+  }
+}
+
+/** The range USFM for one contiguous run: `{2,3} -> "JHN.3.2-3"`, `{5,5} -> "JHN.3.5"`. */
+function runToPassageId(book: string, chapter: string, run: VerseRun): string {
```
jaredhightower-youversion: **suggestion (non-blocking):** Reuse the run→USFM formatter from usfm-ranges.

I noticed `runToPassageId` here is the exact same start===end ternary that already lives inside `buildPassageIds` in usfm-ranges.ts — could we export a per-run formatter from usfm-ranges and reuse it in both places? The machine already imports `collapseVerseRuns` from there, so it's a natural home. (Worth noting `buildPassageIds` itself has no production callers — only its own test — so the exported per-run formatter could become the primitive.)

packages/ui/src/components/use-bible-reader-highlights.ts
```diff
@@ -15,295 +26,225 @@ export type UseBibleReaderHighlightsOptions = {
...
+    if (verse > 0) map[verse] = highlight.color.toLowerCase();
   }
+  return map;
 }
 
-function snapshotOverlay(overlay: HighlightOverlay, verses: number[]): HighlightOverlay {
-  const snapshot: HighlightOverlay = {};
-  for (const verse of verses) {
-    if (verse in overlay) snapshot[verse] = overlay[verse] as string | null;
+function serverColorsEqual(a: ServerColors, b: ServerColors): boolean {
```
jaredhightower-youversion, line 78: **suggestion (non-blocking):** `serverColorsEqual` + the ref bookkeeping look droppable.

I think `serverColorsEqual` and the `lastSentServerColorsRef` bookkeeping can just go — the machine's `reconcileOverlay` is already idempotent for identical server data, and since the hook never reads serverColors back out of the machine, a duplicate `HIGHLIGHTS_UPDATED` doesn't even cause a re-render (`useSelector` only watches overlay). Could we have the effect send `{ type: 'HIGHLIGHTS_UPDATED', serverColors }` unconditionally and drop the ~16 lines of dedup? Bonus: the effect body then matches the shape of the AUTH_CHANGED and SCOPE_CHANGED effects right above it.

packages/ui/src/components/ui/dialog.tsx
```diff
@@ -0,0 +1,68 @@
+import * as React from 'react';
+import * as DialogPrimitive from '@radix-ui/react-dialog';
+
+import { cn } from '../../lib/utils';
+
+const Dialog = DialogPrimitive.Root;
+const DialogTitle = DialogPrimitive.Title;
+const DialogDescription = DialogPrimitive.Description;
+
+function DialogOverlay({
```
jaredhightower-youversion: **suggestion (non-blocking):** Inline `DialogOverlay`.

It's only ever rendered once as `<DialogOverlay />` with no props, so the className merge is never used, and the export isn't imported anywhere either. Could we just inline the `DialogPrimitive.Overlay` directly in `DialogContent` and drop the function plus its export?

packages/hooks/src/useHighlightAuthActions.ts
```diff
@@ -0,0 +1,109 @@
...
+      new ApiClient({
+        appKey: context.appKey,
+        apiHost: context.apiHost,
+        installationId: context.installationId,
+        additionalHeaders: context.additionalHeaders,
+      }),
+    );
+  }, [context?.appKey, context?.apiHost, context?.installationId, context?.additionalHeaders]);
+
+  const hasHighlightsPermission = useCallback(
```
jaredhightower-youversion: **suggestion (non-blocking):** These three `useCallback([])` wrappers can be module-level consts.

None of them capture anything from the hook, so they could just be consts above the hook (e.g. `const hasHighlightsPermission = () => YouVersionPlatformConfiguration.hasPermission(HIGHLIGHTS_PERMISSION);`). That keeps their identity stable for any external consumer while trimming the hook body; `consumeDataExchangeReturn` could even just alias `handleDataExchangeCallback` directly.

packages/ui/src/components/bible-reader-highlights-machine.test.ts
```diff
@@ -0,0 +1,266 @@
...
+
+function startMachine(ref: HighlightServicesRef, scope: HighlightScope = scopeJHN3) {
+  const actor = createActor(bibleReaderHighlightsMachine, {
+    input: { services: ref, scope, flagOn: true, hasAuthProvider: true, isAuthenticated: true },
+  });
+  actor.start();
+  return actor;
+}
+
+async function waitFor(
```
jaredhightower-youversion: **suggestion (non-blocking):** Use `vi.waitFor` instead of the hand-rolled helper.

Vitest already ships `vi.waitFor`, and since none of the callsites actually use the snapshot argument, each one collapses to something like `await vi.waitFor(() => expect(refetch).toHaveBeenCalledTimes(2))`. Same 1s timeout by default, ~11 fewer lines to maintain.

packages/ui/src/components/bible-reader-highlights-machine.test.ts
```diff
@@ -0,0 +1,266 @@
...
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { peekPendingHighlights, readPendingHighlights } from '@/lib/pending-highlight';
+import {
+  bibleReaderHighlightsMachine,
+  type HighlightScope,
+  type HighlightServices,
+  type HighlightServicesRef,
+} from './bible-reader-highlights-machine';
+
+const STORAGE_KEY = 'youversion-platform:pending-highlight';
```
jaredhightower-youversion: **suggestion (non-blocking):** Seed through `appendPendingHighlight` instead of the raw key.

I think we could drop the hardcoded `STORAGE_KEY` here and seed this test with two `appendPendingHighlight(...)` calls instead of the raw `setItem` + JSON blob — it's the same storage state, and it stops the test from duplicating the lib's private key. Bonus: that's the exact path these entries take in production during the redirect scenario, so it reads a bit more faithfully too.

packages/ui/src/components/use-bible-reader-highlights.integration.test.tsx
```diff
@@ -105,3 +117,335 @@ describe('useBibleReaderHighlights — real useHighlights/useApiData', () => {
...
+    // run [5] reverted. Before this fix the WHOLE batch reverted with no
+    // refetch, erasing the persisted 2-3 until the next navigation/write.
+    await waitFor(() => {
+      expect(result.current.highlightedVerses).toEqual({ 2: 'fffe00', 3: 'fffe00' });
+    });
+    await act(async () => {
+      await Promise.resolve();
+    });
+    expect(getHighlights).toHaveBeenCalledTimes(2);
+    consoleError.mockRestore();
```
jaredhightower-youversion: **nitpick (non-blocking):** Redundant `mockRestore` calls.

The `afterEach` at the top of the file already runs `vi.restoreAllMocks()`. The same pattern shows up four times (lines 211/252, 256/301, 305/330, 361/381) — could we drop the restores and make each spy a bare `vi.spyOn(console, 'error').mockImplementation(vi.fn())` since the bindings only exist for the restore?

packages/ui/src/components/use-bible-reader-highlights.test.tsx
```diff
@@ -68,12 +71,22 @@ const defaultOptions = { versionId: 111, book: 'JHN', chapter: '3' };
...
+  // userInfo is persisted (the auth provider does this at sign-in). Seed both so
+  // the authorized-write path is exercised. The auth-flow branches (missing
+  // session/permission) have their own dedicated coverage.
+  YouVersionPlatformConfiguration.saveUserInfo({ id: 'user-1', name: 'Test User' });
+  YouVersionPlatformConfiguration.saveGrantedPermissions(['highlights']);
 });
 
 afterEach(() => {
   setHighlightsLive(HIGHLIGHTS_LIVE);
+  localStorage.clear();
```
jaredhightower-youversion: **nitpick (non-blocking):** Redundant storage clears in `afterEach`.

The `beforeEach` already clears both storages before every test, and Vitest's per-file isolation means nothing leaks to other files. Could we drop those two lines and keep just the `setHighlightsLive` reset?

packages/core/src/__tests__/permissions.test.ts
```diff
@@ -0,0 +1,140 @@
...
+
+      // Only B's grant is present; A's are gone.
+      expect(YouVersionPlatformConfiguration.grantedPermissions).toEqual(['bibles']);
+
+      // And A cannot read them back either.
+      YouVersionPlatformConfiguration.saveUserInfo({ id: 'user-a' });
+      expect(YouVersionPlatformConfiguration.grantedPermissions).toEqual([]);
+    });
+
+    it('same-user merge still accumulates grants', () => {
```
jaredhightower-youversion: **nitpick (non-blocking):** Duplicate coverage.

I think the 'same-user merge still accumulates grants' test can just be deleted — the 'starts empty and merges granted permissions without duplicates' test up top already saves twice for the same user and asserts the merged result (plus dedup and hasPermission on top). This one doesn't add any new coverage inside the scoping block.

packages/ui/src/components/verse-action-popover.test.tsx
```diff
@@ -467,4 +467,45 @@ describe('VerseActionPopover', () => {
...
+      expect(screen.queryByRole('group', { name: 'Highlight colors' })).toBeNull();
+      expect(applyButtons()).toHaveLength(0);
+      expect(clearButtons()).toHaveLength(0);
+
+      // Copy / Share remain.
+      expect(screen.getByText('Copy')).toBeTruthy();
+      expect(screen.getByText('Share')).toBeTruthy();
+    });
+
+    it('still shows the color row by default (highlightsEnabled defaults to true)', () => {
```
jaredhightower-youversion: **nitpick (non-blocking):** Duplicate default-state coverage.

I think the 'still shows the color row by default' test can just be deleted — AC1's 'should display 5 color circles' already renders with the same default props and would catch a default flip of `highlightsEnabled`. The flag-off test above it carries all the new coverage here.

packages/ui/src/components/use-bible-reader-highlights.auth-flow.integration.test.tsx
```diff
@@ -0,0 +1,436 @@
...
+
+describe('highlight auth flow — data-exchange return', () => {
+  it('applies the pending highlight on a granted return (async session hydration)', async () => {
+    signedIn = false;
+    setLocation(
+      'https://host.example/read?data_exchange_status=granted&granted_permissions=highlights',
+    );
+    // Pre-stash a pending highlight as the confirm path would have.
+    sessionStorage.setItem(
+      'youversion-platform:pending-highlight',
```
jaredhightower-youversion: **suggestion (non-blocking):** Seed through the real writers instead of the raw key.

These three tests hand-build the sessionStorage entry with the raw 'youversion-platform:pending-highlight' key, but the module under test already exports `stashPendingHighlight`/`appendPendingHighlight` (and this file imports `readPendingHighlights` from it). Seeding through the real writer would drop the hardcoded key and the hand-copied entry shape — right now the storage format lives in three files and they'll drift if it ever changes.

jaredhightower-youversion (commented):
Ponytail pass (over-engineering only) — all non-blocking. Correctness was already well covered by the existing threads.

packages/core/src/data-exchange.ts
```diff
@@ -0,0 +1,148 @@
...
+   */
+  async updateToken(permissions: string[], lat?: string): Promise<string> {
+    const appKey = this.client.config.appKey;
+    if (!appKey) {
+      throw new Error('App key is required to request a data exchange token.');
+    }
+
+    const response = await this.client.post<unknown>(
+      `/data-exchange/token`,
+      { requested_permissions: [...permissions].sort() },
```
jaredhightower-youversion: **question (non-blocking):** Is there a server contract that needs sorted permissions?

`updateToken` copies and sorts the permissions before sending — if there's no ordering contract, I think plain `{ requested_permissions: permissions }` would do; the only caller passes a single permission anyway, and dropping the sort also lets us drop the defensive spread-copy.

packages/core/src/permissions.ts
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * The param may repeat and each value may pack several permissions separated by
+ * a comma or whitespace (mirrors the Swift SDK's split on `,`/` `). Returns a
+ * de-duplicated list, order-preserving on first appearance.
+ */
+export function parseGrantedPermissions(params: URLSearchParams): string[] {
+  const seen = new Set<string>();
+  for (const value of params.getAll('granted_permissions')) {
+    for (const part of value.split(/[,\s]+/)) {
+      const trimmed = part.trim();
```
jaredhightower-youversion: **nitpick (non-blocking):** `part.trim()` is a no-op here.

Since the split regex `/[,\s]+/` already eats all whitespace as part of the delimiter, this could collapse to just `if (part) seen.add(part);`.

packages/ui/src/lib/pending-highlight.ts
```diff
@@ -0,0 +1,164 @@
...
+ * @param now Injectable clock for tests; defaults to `Date.now()`.
+ */
+export function appendPendingHighlight(pending: PendingHighlight, now: number = Date.now()): void {
+  const existing = readStoredPending(now).entries;
+  const claimedVerses = new Set(pending.verses);
+  const merged: PendingHighlight[] = [];
+  for (const entry of existing) {
+    const remaining = entry.verses.filter((verse) => !claimedVerses.has(verse));
+    if (remaining.length === 0) continue;
+    merged.push(remaining.length === entry.verses.length ? entry : { ...entry, verses: remaining });
```
jaredhightower-youversion: **nitpick (non-blocking):** Micro-optimization that can't be observed.

The entries were just freshly parsed from JSON and get immediately stringified back, so nothing can observe whether we reused `entry` or copied it. Could we just always do `merged.push({ ...entry, verses: remaining })`?

docs/highlight-flow-statechart.md
```diff
@@ -0,0 +1,109 @@
...
+      checkQueue --> w_idle: queue empty
+    }
+  }
+```
+
+_(`ENQUEUE`, `AUTH_CHANGED`, `HIGHLIGHTS_UPDATED`, and `SCOPE_CHANGED` are handled
+without leaving the current state: they update context and let the `always`
+guards re-route. `HIGHLIGHTS_UPDATED` also runs the overlay reconcile.)_
+
+## The "vapor" fix
```
jaredhightower-youversion, line 96: **suggestion (non-blocking):** Point at the code comment instead of duplicating it.

The "vapor fix" section here duplicates the block comment at the top of bible-reader-highlights-machine.ts almost word for word — root cause, fix, and trade-off. Could we replace it with a one-liner like "See the 'vapor fix' block comment at the top of bible-reader-highlights-machine.ts for the rationale"? That way the prose next to `reconcileOverlay` stays the single source of truth and the two copies can't drift. (Related: that code comment references docs/adr / CONTEXT.md for the mermaid diagram, but the diagram actually lives here — worth fixing the pointer in the same pass.)

.changeset/highlight-auth-flow.md
```diff
@@ -0,0 +1,19 @@
...
+- **Hooks**: new `useHighlightAuthActions` exposing the one-fell-swoop sign-in (requesting `highlights`), the just-in-time data-exchange redirect, the permission-cache reads/invalidation, and the data-exchange return handler.
+- **UI**: `useBibleReaderHighlights` now runs the state machine — pending highlights persist to `sessionStorage` (~10-minute expiry) to survive the redirect round-trip and apply automatically on a granted return; a just-in-time permission confirm dialog (`HighlightPermissionDialog`, copy matched to the native SDK) gates the data-exchange grant. Write failures route by status: 401/403 invalidates the cache, keeps the pending highlight, and re-prompts; 5xx/network reverts the optimistic overlay and discards. Apply/remove writes are serialized through a FIFO queue with per-verse ownership so overlapping operations settle to the last-issued state. Copy/share-only behavior when no auth provider is configured is unchanged.
+
+Deferred follow-ups (documented at each code site in `bible-reader-highlights-machine.ts`, accepted for dark launch):
```
jaredhightower-youversion: **nitpick (non-blocking):** Internal tracking in a consumer-facing changelog.

I think the "Deferred follow-ups" section and the trailing Note could just be deleted from this changeset — it's internal known-issue tracking that's already documented at each code site in bible-reader-highlights-machine.ts, and the PR-288 supersession is covered by the xstate changeset. Since this ships in the published CHANGELOG and the flow is still behind HIGHLIGHTS_LIVE, consumers won't get anything actionable from the auth-expiry/resume-write internals (item 3 is even struck through and already resolved).

.changeset/xstate-highlights-flow.md
```diff
@@ -0,0 +1,12 @@
+---
+'@youversion/platform-react-ui': patch
+---
+
+Rewrite the BibleReader highlights flow as an explicit xstate v5 statechart (YPE-1034 / PR-288, still behind the internal `HIGHLIGHTS_LIVE` flag). `useBibleReaderHighlights` is now a thin adapter over `bibleReaderHighlightsMachine`; the machine (authored with `setup()` and named guards/actions/actors, so it is Stately-visualizable) owns the whole flow — optimistic overlay, serialized writes with per-verse ownership, reconcile, the auth flow, and the resume-on-return path. A mermaid statechart lives in `docs/highlight-flow-statechart.md`. All previously-shipped invariants are preserved (writes serialized, per-verse ownership tokens, per-sub-write settlement, apply collapses to ranges / remove DELETEs per verse, one refetch per settled batch, scope-change drops the overlay, `live` gates rendering/fetching).
```
jaredhightower-youversion: **nitpick (non-blocking):** Trim the invariant list.

"All previously-shipped invariants are preserved." already carries the signal — the parenthetical list of things that didn't change is internal detail that mostly repeats the earlier sentence about what the machine owns. Since changesets end up in the consumer-facing CHANGELOG, could we keep it to the short sentence?

Dustin-Kelley (dismissed):

packages/core/src/SignInWithYouVersionResult.ts
cameronapak: Done in bb528c1: dropped the `permissions` field, its ctor param, and the `grantedPermissions` threading through `extractSignInResult`. The permission cache on `YouVersionPlatformConfiguration` is the single source of truth; the callback still seeds it via `saveGrantedPermissions`. Removed the changeset sentence about the field too.

packages/core/src/YouVersionPlatformConfiguration.ts
cameronapak: Done in bb528c1: deleted `setGrantedPermissions` and its unit test. Verified no callers anywhere in the stack (285/286/294).

packages/core/src/data-exchange.ts
cameronapak: Agreed on the dup, but this exact dedupe already lands in stack 4/4: #294 commit 700ba155 extracts `core/src/auth-token.ts` with `resolveAuthToken(lat, action)` and moves both `DataExchangeClient` and `HighlightsClient` onto it. Doing it here as well would only create a rebase conflict with that commit, so 285 stays as-is and 294 carries the fix.

packages/ui/src/components/bible-reader-highlights-machine.ts
cameronapak: Done in bb528c1: `usfm-ranges` now exports `formatPassageId(book, chapter, run)` and both `buildPassageIds` and the machine use it (`runToPassageId` deleted). Added direct unit tests for the new export.

packages/ui/src/components/use-bible-reader-highlights.ts
cameronapak: Holding this one for the stack: 4/4 (#294) builds on these refs. It adds `parsedServerColorsRef` using `serverColorsEqual` to keep `serverColors` reference-stable for downstream memos, alongside the existing send-dedup ref. Dropping them here would break those commits. Happy to revisit post-stack if it still looks redundant then.

packages/ui/src/components/ui/dialog.tsx
cameronapak: Done in bb528c1: inlined the overlay into `DialogContent` and dropped the function and export. Nothing imported it and the className merge was dead.

packages/hooks/src/useHighlightAuthActions.ts
cameronapak: Done in bb528c1: all three are module-level consts above the hook now. Return shape unchanged, and the `dataExchangeClient` memo above was left untouched so 294's edit there rebases clean.

packages/ui/src/components/bible-reader-highlights-machine.test.ts
cameronapak: Done in bb528c1: helper deleted, all 10 callsites on `vi.waitFor` with `expect()` predicates (same 1s timeout as the helper). Confirmed none used the snapshot arg.

packages/ui/src/components/bible-reader-highlights-machine.test.ts
cameronapak: Done in bb528c1: the test seeds via two `appendPendingHighlight` calls now, and the `STORAGE_KEY` const is gone from the test file.

packages/ui/src/components/use-bible-reader-highlights.integration.test.tsx
cameronapak: Done in bb528c1: dropped the four restores (verified each was the last statement of its test) and made the spies bare.

packages/ui/src/components/use-bible-reader-highlights.test.tsx
cameronapak: Done in bb528c1: `afterEach` keeps only the `setHighlightsLive` reset.

packages/core/src/__tests__/permissions.test.ts
cameronapak: Done in bb528c1: deleted. Confirmed the first test covers the same-user double save with merge and dedup assertions.

packages/ui/src/components/verse-action-popover.test.tsx
cameronapak: Done in bb528c1: deleted. The AC1 test renders the same default props and its 5-circle assertion fails if `highlightsEnabled` ever defaults to false, so the coverage survives.

packages/ui/src/components/use-bible-reader-highlights.auth-flow.integration.test.tsx
cameronapak: Done in bb528c1: all three seed via `stashPendingHighlight` now (each seeded a single valid entry, so the whole-list stash path matches exactly). Heads up: #294 adds one more raw-key seed in its new tests; that aligns when it rebases onto this.

packages/core/src/data-exchange.ts
cameronapak: No ordering contract that we know of; the sort was defensive determinism only. Dropped in bb528c1: `{ requested_permissions: permissions }`.

packages/core/src/permissions.ts
cameronapak: Done in bb528c1: collapsed to `if (part) seen.add(part);`.

packages/ui/src/lib/pending-highlight.ts
cameronapak: Done in bb528c1: always copies now.

docs/highlight-flow-statechart.md
cameronapak: Done in bb528c1: the doc section is now a one-line pointer to the machine's block comment (which sits next to `reconcileOverlay`). Also fixed that comment's stale mermaid pointer to reference `docs/highlight-flow-statechart.md`.

.changeset/highlight-auth-flow.md
cameronapak: Done in bb528c1: the Deferred follow-ups section and trailing Note are gone; the code sites keep the tracking.

.changeset/xstate-highlights-flow.md
cameronapak: Done in bb528c1: parenthetical dropped, the sentence stands alone. Left the Behavior changes list untouched since #294 appends to it.

packages/hooks/src/useHighlights.ts
```diff
@@ -54,21 +54,21 @@ export function useHighlights(
...
-      await highlightsClient.deleteHighlight(passageId, deleteOptions);
-      refetch();
-    },
-    [highlightsClient, refetch],
+    (passageId: string, deleteOptions: DeleteHighlightOptions): Promise<void> =>
+      highlightsClient.deleteHighlight(passageId, deleteOptions),
+    [highlightsClient],
   );
 
   return {
```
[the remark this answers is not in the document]
cameronapak: No. Highlights isn't used yet.

jaredhightower-youversion (dismissed):
Approved but there is one small comment from greptile.

packages/hooks/src/useHighlights.ts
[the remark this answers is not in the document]
cameronapak: Fair catch — main (published 2.3.0) auto-refetches and this PR stops that on a public hook. Took both mitigations in 353d526: JSDoc on `createHighlight`/`deleteHighlight` return-type members (so the contract surfaces in editor tooltips: no auto-refetch, call `refetch()` after the batch settles) and the hooks entry in the refetch-coalescing changeset flipped patch → minor, with the changeset wording updated ("sole in-repo consumer" + explicit note that direct consumers must refetch themselves). No known external consumers today, so minor over major: the auto-refetch was undocumented behavior, and the highlights surface is pre-GA behind HIGHLIGHTS_LIVE.

jaredhightower-youversion (approved):

PR #286, opened 2026-07-10
Checkmark marks the active highlight swatch (YPE-1034) (3/4)

cameronapak:
## Summary

The active highlight swatch renders a 24px checkmark (matching iOS) instead of an X. This PR's original recent-colors feature was dropped from the SDK entirely — the `/v1/highlights/recent-colors` contract is still changing and must not ship (per bbelyeu; main removed the API in #291) — leaving the checkmark as the surviving slice.

Stack: 1/4 #283 → 2/4 #285 → **3/4 (this PR)** → 4/4 #294 · restored from #288

## Changes

1. Active and remove swatches render a `CheckIcon` (24px, iOS-matching path) instead of the X.
2. ADR-005 records the active-swatch icon decision.

## Test plan

- `pnpm build && pnpm typecheck && pnpm lint && pnpm test` at this branch: all green (core 315 + checkmark suite; ui 258 / hooks 284).
- `verse-action-popover.test.tsx` asserts the checkmark SVG path on active/remove swatches.

<h3>Greptile Summary</h3>

Replaces the X icon on active highlight swatches with a 24 px checkmark, matching iOS parity (`platform-sdk-swift #179`). Behavior is unchanged — tapping a checkmark swatch still removes the highlight; only the indicator icon changes.

- `packages/ui/src/components/icons/check.tsx`: new `CheckIcon` with `aria-hidden` default and `{...props}` spread, consistent with other icon files.
- `packages/ui/src/components/verse-action-popover.tsx`: swaps `XIcon` → `CheckIcon` on active swatches; renames the `showX` prop to `showRemove` for clarity.
- Tests, changeset, and ADR note are all included and accurate.

<h3>Confidence Score: 5/5</h3>

Safe to merge — the change is a pure icon swap with no behavioral or API surface changes.

The change is minimal and well-contained: a new icon component, a single prop rename, and two targeted tests that directly verify the checkmark path. Behavior (highlight removal on tap, aria-labels, flag gating) is untouched. The new icon correctly sets aria-hidden before the props spread, changeset and ADR are both present, and no new i18n strings or public API exports were added.

No files require special attention.

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
| packages/ui/src/components/icons/check.tsx | New CheckIcon SVG component with aria-hidden default and props spread — follows existing icon conventions; improves on XIcon which lacked aria-hidden. |
[...]

PR #294, opened 2026-07-20
Final wiring of all things highlights API and auth-guard, with fix for highlight not deleting immediately (visually)

cameronapak:
# [PLZ WATCH VIDEO](https://cap.so/s/4rmk47t6nt89gy1) ▶️

## Summary

Server-backed BibleReader highlights go live: the dark-launch flag flips on after a final round of auth hardening, a fix for successful deletes being misread as failures, and Swift-parity UX corrections. Last PR of the YPE-1034 stack (#283 → #285 → #286 → this one); the consolidated changelog announces the launch and the removal of the accidentally shipped localStorage-only highlights.

## Changes

1. Highlights are enabled by default (`HIGHLIGHTS_LIVE` on) — signed-in readers' highlights persist to their YouVersion account; six stacked changesets are consolidated into one launch changelog.
2. Removing a highlight no longer flashes back before disappearing — the core `ApiClient` treated empty-body 2xx JSON responses (the real DELETE shape) as failures, so the machine reverted the successful removal; regression tests now drive the real client through a stubbed `fetch`.
3. A data-exchange grant only applies to the user who initiated it — initiator recorded at redirect, read-and-cleared at callback, any mismatch fails closed; `signIn` also clears stale permission stashes so an abandoned flow can't leak grants across users.
4. A resumed pending write that fails 401/403 re-prompts for permission instead of silently reverting the reader's tap.
5. Sign-in matches the Swift SDK: `appName` / `signInPromptMessage` props, Platform wordmark, comma-joined `requested_permissions` wire format, and a permission cache seeded from token scope so a highlight tap doesn't re-prompt right after OAuth.
6. Dark mode matches the Swift SDK: highlight fills render at 0.3 opacity (1.0 in light mode), verse numbers paint white over dark-mode fills, and popover swatches preview the dimmed color a highlight will actually apply; fills also get subtly rounded corners with per-line-fragment rounding at wraps (deliberate web-native divergence from Swift's square fills); the active swatch shows a checkmark instead of an X.
7. Hardening and cleanup: xstate pinned to exactly 5.32.4 (supply-chain cooldown), review findings across all three packages (color row gated on interactivity, stable `serverColors`/`refetch` identities, shared `useApiClient`), the logo's accessible label is a required localized prop, the pending-permissions stash stores `string[]`, and personal workflow notes are stripped from AGENTS.md.

8. One-shot sign-in grants survive the server's URL encoding: `parseGrantedPermissions` now also accepts bracket-array keys (`granted_permissions[]`, indexed forms) that the auth server emits, on both the sign-in callback and the data-exchange return — previously the echoed grant was silently discarded and the permission dialog re-prompted after a successful sign-in.

9. StrictMode-proof auth: the OAuth code exchange and the token refresh are each shared across concurrent invocations, so a duplicate's 4xx can no longer clear the session or the just-seeded grants (previously the demo's `<StrictMode>` double-invoke could wipe a fresh sign-in or grant cache).

10. One-shot sign-in survives a silent server: a live instrumented capture (2026-07-23) showed the web callback carries **no** `granted_permissions` echo on either hop and the token scope is `profile openid` only — the SDK had zero evidence of the consent the user just gave. The permissions requested at `signIn` are now stashed (state-bound, fail-closed) and optimistically seeded as granted on a successful exchange; an over-optimistic seed self-corrects through the existing 401/403 → permission re-prompt path. The durable fix is server-side (echo the grant on the web callback) and is being filed with Platform.

11. Verse-number labels and footnote icons inherit the verse text color over highlight fills in both themes, so saturated light-mode fills no longer clash with muted gray glyphs (deliberate divergence from Swift, which only recolors the label in dark mode).

12. Reviewer simplification pass applied in full (12 comments, −286 lines: shared highlight test fixtures, dead helper/config removal, duplicate test deleted) and `getHttpStatus` now recognizes plain `{status}` objects, not just `Error` instances.

13. Mobile: the verse-action popover caps at the viewport width (with live tracking across mid-open resizes) and the swatch row scrolls horizontally with edge fades that appear only where content is hidden; Copy/Share stay pinned. Verified live at 300px.

14. Fill opacity churn, net zero: a same-day design decision briefly moved light-mode fills to 30% opacity and was then walked back — final state is the original 1.0 light / 0.3 dark (Swift parity). Kept in history as commits cf6a291/959f2ca and revert 7b91105.

**Start here:** change 3 — auth-boundary logic; the fail-closed path is the part worth adversarial review.

## Test plan

- `pnpm build && pnpm typecheck && pnpm lint && pnpm test`: all green (1014 tests — core 369 / hooks 289 / ui 356)
- Vapor regression suite fails on the reverted fix, passes with it (fetch-stub end-to-end through the real client, machine, and adapter)
- Storybook integration tests run in CI

**Needs manual check:** one live confirmation in the vite demo — remove a highlight while signed in and confirm no flash. The one-shot flow's root cause was pinned down via a breadcrumb-instrumented live run (server returns no grant evidence; see change 10) and the fix is regression-tested against that exact captured shape. Live-verified 2026-07-23: signed-out tap → sign-in → OAuth → highlight applies with no second dialog.

<h3>Greptile Summary</h3>

This is the final PR in the YPE-1034 stack, flipping server-backed BibleReader highlights to live. It ships a substantial auth hardening pass alongside the feature launch: the empty-body DELETE flash fix, StrictMode-safe code-exchange and token-refresh deduplication, initiator-bound data-exchange grant safety, bracket-array permission parsing, and optimistic permission seeding for the web flow's missing server grant echo.

- **Auth boundary (change 3 + 9):** the data-exchange initiator is recorded before the redirect and cleared on return; any user mismatch or missing initiator fails closed. Code-exchange and token-refresh deduplication prevent StrictMode double-invocations from clearing tokens the winning run just seeded.
- **DELETE flash fix (change 2):** `ApiClient` now reads the response body as text first and treats an empty body as \"no data\", so a `200 application/json` with no payload no longer throws and surfaces a successful delete as a failure.
- **Permission seeding (changes 8 + 10):** `parseGrantedPermissions` accepts bracket-array keys (`granted_permissions[]`, `granted_permissions[0]`), and requested permissions are optimistically seeded on callback (state-bound, self-correcting via 401/403 → re-prompt) to work around the server returning no grant echo on the web flow.

<h3>Confidence Score: 4/5</h3>

Safe to merge except for the light-mode fill opacity constant, which doesn't match the product decision documented in the PR description.

The auth hardening, DELETE flash fix, and StrictMode deduplication are all well-designed and regression-tested. The one concrete gap is that `HIGHLIGHT_FILL_OPACITY_LIGHT` is still `1.0` even though the PR description's product decision (item 14) explicitly changed light mode to 30% — matching dark mode — and says swatches should preview the dimmed fill in both modes. The ADR update was written before that decision landed and still documents the pre-decision split. Everything else in the 63-file change looks correct.

packages/ui/src/components/verse.tsx (fill opacity constant) and packages/ui/src/components/verse-action-popover.tsx (swatch preview in light mode)

| Filename | Overview |
|----------|----------|
| packages/core/src/Users.ts | Major auth hardening: StrictMode-safe code-exchange deduplication, state-bound permission stashes, optimistic permission seeding with self-correcting 401/403 path, and stale-stash cleanup at signIn start. Well-structured and test-covered. |
| packages/core/src/data-exchange.ts | Initiator-bound grant safety: initiator is recorded at redirect start, read-and-cleared on return, and any mismatch (different user, missing initiator) degrades to failure. Security design is solid and regression-tested. |
| packages/core/src/client.ts | Adds `getHttpStatus` with duck-type check (accepts any object with a numeric `status` property, not just Error instances), and fixes empty-body 2xx handling so DELETE responses no longer surface as JSON parse failures. |
| packages/core/src/permissions.ts | Adds bracket-array key support (`granted_permissions[]`, `granted_permissions[0]`) via regex to handle PHP/Rails server encoding; extracts `parsePermissionList` for token-scope parsing. Correctly handles all observed server formats. |
| packages/ui/src/components/bible-reader-highlights-machine.ts | Removes the `reprompt` field from `WriteOp` — all apply-kind writes (user-initiated and resume-applied) now re-prompt on 401/403, fixing silent drop of resumed taps. Adds `signedOut` guard to auto-dismiss the permission dialog on sign-out. |
| packages/ui/src/components/verse.tsx | Introduces theme-aware fill opacity constants (1.0 light / 0.3 dark), but `HIGHLIGHT_FILL_OPACITY_LIGHT = 1.0` contradicts the PR description's stated product decision that both modes should render at 30%. |
| packages/core/src/YouVersionPlatformConfiguration.ts | Adds `dataExchangeInitiator` storage (save/get/clear) and wires `clearDataExchangeInitiator` into `clearAuthTokens` so a sign-out always drops a pending initiator. |
| packages/hooks/src/internal/useApiClient.ts | New shared hook that centralizes ApiClient construction for all per-service hooks, reducing duplication. Optional mode (`{ optional: true }`) returns null instead of throwing for auth-tolerant callers. |
| packages/hooks/src/useApiData.ts | Adopts the `useEvent`-style `fetchFnRef` pattern to keep `fetchFn` out of `fetchData`'s deps, stabilizing `refetch` identity across renders without stale-closure risk. |
| packages/ui/src/components/verse-action-popover.tsx | Adds horizontal scroll with per-edge fade masks (pure `computeScrollFade` function, ResizeObserver-tracked), viewport-width cap, corrected auto-focus on pointer-open, and theme-aware swatch previews. |
| packages/ui/src/components/use-bible-reader-highlights.ts | Adds `highlightsInteractive` flag and a reference-stable `serverColors` memoization (content-equality comparison via ref) to prevent spurious verse-style re-sweeps on no-op refetches. |
| packages/hooks/src/useHighlightAuthActions.ts | Saves the data-exchange initiator before the async token mint and clears it on error, ensuring any mid-await user change can't stamp the new session as the original initiator. |

<sub>Reviews (14): Last reviewed commit: ["revert(ui): restore full-strength light-..."](https://github.com/youversion/platform-sdk-react/commit/7b911050ab6f1d6d6efff7bb03073b906b624974) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=45746502)</sub>


packages/core/src/permissions.ts
```diff
@@ -14,10 +14,43 @@
...
+  for (const [key, value] of params) {
+    if (isGrantedPermissionsKey(key)) {
+      values.push(value);
+    }
+  }
+  return mergePermissionValues(values);
+}
+
+/** Matches `granted_permissions`, `granted_permissions[]`, or `granted_permissions[<index>]`. */
+function isGrantedPermissionsKey(key: string): boolean {
```
jaredhightower-youversion: **suggestion (non-blocking):** Single-use helper could be one regex.

I think we could drop the `isGrantedPermissionsKey` helper — it's only called once, and a single regex like `/^granted_permissions(\[\d*\])?$/.test(key)` inside the loop covers all three key shapes identically. The big doc comment above already explains which keys we accept, so the helper's JSDoc isn't pulling extra weight.

jaredhightower-youversion (commented):
Simplification pass — a few candidates to delete or shrink; all non-blocking.

packages/core/src/SignInWithYouVersionPKCE.ts
```diff
@@ -69,11 +69,12 @@ export class SignInWithYouVersionPKCEAuthorizationRequestBuilder {
...
     // YouVersion data-exchange permissions (e.g. `highlights`) are intentionally
-    // NOT OIDC scopes. They ride alongside the standard scopes as a repeatable
-    // `requested_permissions[]` query param and are authorized via a separate
-    // per-app ACL rather than the token's scope claim.
-    for (const permission of permissions ?? []) {
-      queryParams.append('requested_permissions[]', permission);
+    // NOT OIDC scopes. They ride alongside `scope` as a single comma-joined
+    // `requested_permissions` query param (Swift/Kotlin wire format) and are
+    // authorized via a separate per-app ACL rather than the token's scope claim.
+    const permissionsValue = [...(permissions ?? [])].filter(Boolean).sort().join(',');
```
jaredhightower-youversion: **nitpick (non-blocking):** `.filter(Boolean)` is dead code here.

`permissions` is typed as `SignInWithYouVersionPermissionValues[]`, which can't contain falsy values, and `scopeValue()` right above doesn't do this kind of filtering for scopes either. Could we simplify to `const permissionsValue = [...(permissions ?? [])].sort().join(',');`?

packages/ui/src/assets/youversion-platform-logo/Contents.json
```diff
@@ -0,0 +1,22 @@
+{
```
jaredhightower-youversion: **suggestion (non-blocking):** The whole imageset folder looks like dead config.

I think we can drop `src/assets/youversion-platform-logo/` entirely — this Contents.json is an Xcode asset-catalog manifest nothing in this web package reads, tsup doesn't process SVGs, and the folder never ships since only `dist` is published. The path data already lives in `WORDMARK_PATH` with both fills in `FILL`. If it's kept as Swift-parity reference, could the doc comment in `youversion-platform-logo.tsx` just name the Swift `YouVersionPlatformLogo.imageset` instead? (If deleted, that doc comment needs updating so it doesn't point at a nonexistent path.)

packages/core/src/__tests__/Users.test.ts
```diff
@@ -256,6 +309,248 @@ describe('YouVersionAPIUsers', () => {
       saveAuthDataSpy.mockRestore();
     });
 
+    it('unions stashed early grants with token scope when final URL omits them', async () => {
+      const mockTokens = {
```
jaredhightower-youversion: **suggestion (non-blocking):** Same 8-line `mockTokens` literal pasted into four tests.

I noticed the same mockTokens literal (long JWT included) is pasted into all four new callback tests with only the scope string changing — could we pull it into a small `makeTokens(scope)` helper? That'd trim ~28 lines and the existing test at line 212 could reuse it too. The duplication actually goes further — the localStorage mock, fetch mock, and atob payload are near-identical across the same tests, so a single `setupCallbackFlow({ scope, pendingGrant })` helper would shrink each test to its distinguishing setup plus assertions.

packages/core/src/__tests__/Users.test.ts
```diff
@@ -95,7 +95,29 @@ describe('YouVersionAPIUsers', () => {
...
       await YouVersionAPIUsers.signIn('https://example.com/callback', ['profile'], ['highlights']);
 
-      expect(mocks.window.location.href).toContain('requested_permissions%5B%5D=highlights');
+      expect(mocks.window.location.href).toContain('requested_permissions=highlights');
+
+      vi.restoreAllMocks();
+    });
+
+    it('clears any stale pre-code granted-permissions stash from a prior abandoned flow', async () => {
+      vi.spyOn(crypto, 'getRandomValues').mockImplementation((array: ArrayBufferView) => {
```
jaredhightower-youversion: **nitpick (non-blocking):** Third verbatim copy of the crypto stub block.

The new stale-stash test copies the same 11-line getRandomValues/digest/btoa stub the two signIn tests above already have. Could we pull that into a tiny `stubSignInCrypto()` helper (or the signIn describe's beforeEach)? Relatedly, the trailing `vi.restoreAllMocks()` calls could move into one `afterEach`, which also guarantees restoration when an assertion fails mid-test.

packages/ui/src/components/use-bible-reader-highlights.dom-vapor.test.tsx
```diff
@@ -0,0 +1,161 @@
...
+  type YouVersionUserInfo,
+} from '@youversion/platform-core';
+import { YouVersionAuthContext, YouVersionContext } from '@youversion/platform-react-hooks';
+import type { ReactNode } from 'react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { HIGHLIGHTS_LIVE, setHighlightsLive } from '@/lib/feature-flags';
+import { useBibleReaderHighlights } from './use-bible-reader-highlights';
+import { Verse } from './verse';
+
+function collection(data: Highlight[]): Collection<Highlight> {
```
jaredhightower-youversion: **suggestion (non-blocking):** Shared test fixtures are copy-pasted across ~5 highlight test files.

`collection()`, `mockUserInfo`, `deferred()`, and the `Providers` wrapper are byte-identical across the three new vapor test files, and the older highlights tests carry the same copies. Could we pull them into `packages/ui/src/test/` next to the existing `utils.ts`? `Providers` just needs an optional `userInfo` prop to cover the signed-out variant.

packages/ui/src/components/verse-action-popover.test.tsx
```diff
@@ -506,6 +527,81 @@ describe('VerseActionPopover', () => {
         expect(btn.querySelector('svg')).toBeNull();
       });
     });
+
+    it('tapping the checkmark swatch still removes the highlight', () => {
```
jaredhightower-youversion: **suggestion (non-blocking):** This test is an exact rerun of an earlier one.

'tapping the checkmark swatch still removes the highlight' has the same props, click, and assertion as 'should call onClearHighlight with color when X circle clicked' above, with nothing checkmark-specific in it — I think it can just be deleted. The checkmark itself is already covered by the path-prefix test right before it. (If deleted, the surviving test's "X circle" name is stale since the icon changed — maybe rename to "clear swatch".)

packages/ui/src/components/verse.test.tsx
```diff
@@ -587,6 +589,139 @@ describe('Verse.Html - Intro Chapter Footnotes', () => {
...
+describe('Verse.Html - Rounded highlight fill (static structural CSS)', () => {
+  // The rounded corners / clone / padding are structural styles that live in the
+  // core stylesheet (`bible-reader.css`), not the imperative paint path (which
+  // only sets colors). They are applied to the base `.yv-v` rule so they are
+  // STATIC — present whether or not a verse is highlighted — which is what keeps
+  // applying/removing a fill from reflowing text (no layout shift). jsdom doesn't
+  // load that external sheet, so we assert against the CSS source directly.
+  // Resolve from cwd so it works whether the suite runs from the ui package
+  // (the filtered command) or the repo root (turbo).
+  const cssPath = [
```
jaredhightower-youversion: **suggestion (non-blocking):** cwd probe can be one `import.meta.dirname` resolve.

The test file's location is fixed relative to the CSS, so `readFileSync(resolve(import.meta.dirname, '../../../core/src/styles/bible-reader.css'), 'utf8')` works from both the filtered ui run and the turbo root run. That drops the candidate array, `existsSync`, `.find`, and the manual throw — `readFileSync` already fails loudly if the file moves.

packages/ui/src/components/YouVersionProvider.test.tsx
```diff
@@ -46,6 +46,29 @@ describe('UI YouVersionProvider', () => {
...
+      <YouVersionProvider
+        appKey="test-key"
+        appName="SDK Demo"
+        signInPromptMessage="Save your highlights to your YouVersion account."
+      >
+        <div data-testid="child">hello</div>
+      </YouVersionProvider>,
+    );
+
+    await vi.waitFor(() => {
```
jaredhightower-youversion: **nitpick (non-blocking):** `vi.waitFor` unnecessary — effect already committed.

`render()` wraps in `act()`, so the config-mirroring effect has already run by the time it returns; plain assertions pass and match the immediate-assert style of every other test in this file. The `await import('@youversion/platform-core')` on line 50 could also become a static import since core isn't mocked here.

packages/ui/src/components/use-bible-reader-highlights.dom-vapor.test.tsx
```diff
@@ -0,0 +1,161 @@
...
+      highlightedVerses={api.highlightedVerses}
+      onVerseSelect={() => undefined}
+    />
+  );
+}
+
+describe('vapor flash — real Verse.Html DOM paint (MutationObserver on style)', () => {
+  it('the verse-2 background is never repainted to yellow after the optimistic unpaint', async () => {
+    const withRow = () => collection([{ version_id: 111, passage_id: 'JHN.1.2', color: 'fffe00' }]);
+    const held = deferred<Collection<Highlight>>();
```
jaredhightower-youversion: **nitpick (non-blocking):** `deferred()` used only as a never-settling promise.

`held.resolve` is never called, so this is really just `new Promise<Collection<Highlight>>(() => {})` — dropping the helper makes the "refetch stays in flight forever" intent more obvious.

packages/ui/src/components/use-bible-reader-highlights.dom-vapor.test.tsx
```diff
@@ -0,0 +1,161 @@
...
+afterEach(() => {
+  vi.restoreAllMocks();
+  setHighlightsLive(HIGHLIGHTS_LIVE);
+  localStorage.clear();
+  sessionStorage.clear();
+});
+
+function Reader({ removeRef }: { removeRef: { current: (() => void) | null } }) {
+  const [selected, setSelected] = useState<number[]>([2]);
+  const sectionRef = useRef<HTMLDivElement>(null);
```
jaredhightower-youversion: **nitpick (non-blocking):** `sectionRef` never read.

It's passed to `Verse.Html` but assertions all go through `container.querySelector` and the MutationObserver — the `useRef` line, the `ref` prop, and `useRef` in the import can all go.

packages/ui/src/components/use-bible-reader-highlights.strictmode-vapor.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+});
+
+afterEach(() => {
+  vi.restoreAllMocks();
+  setHighlightsLive(HIGHLIGHTS_LIVE);
+  localStorage.clear();
+  sessionStorage.clear();
+});
+
+type Frame = { hv: Record<number, string>; scopeMatch: boolean };
```
jaredhightower-youversion: **nitpick (non-blocking):** `Frame` wrapper is vestigial.

`scopeMatch` is hardcoded `true` and nothing asserts on it (the filter only looks at `hv[2]`). Could we type frames as `Record<number, string>[]` and push `api.highlightedVerses` directly? The file header's "machineScope/scope equality" claim is stale too.

jaredhightower-youversion (approved):
LGTM!! Awesome job on this game. It's been a journey, but we've gotten to this point. I did watch the video. The small little nuances with the thing getting cut off: not sure if you were able to resolve that. If not, no worries. A small follow-up PR should be able to resolve that.

Dustin-Kelley (approved):
🔥

PR #301, opened 2026-07-27
feat(ui)!: make BibleReader highlights opt-in via enableHighlights

cameronapak:
[Structure outline](https://cloud.humanlayer.com/artifacts/019fa46b-90ff-7741-914f-76803f03982b) | [PR Walkthrough (alpha)](https://cloud.humanlayer.com/artifacts/019fa4b4-75a2-7a95-9be2-332976be4613)

> **Behavior change.** Highlights shipped enabled by default in 2.4.0. After this PR they are **off unless the host asks for them**. Migration is one prop: add `enableHighlights` to `BibleReader.Root`. Controlled-mode hosts (passing `highlights`) are unaffected.

Any host that mounted `BibleReader.Root` inside a `YouVersionProvider` with `includeAuth` got server-backed highlights whether or not they wanted them: five color swatches in the verse action bar, `/v1/highlights` traffic, consent dialogs, and the possibility of a full-page OAuth redirect that navigates the user out of the host's app. The host never asked for any of it, and had no way to decline short of not using `BibleReader`.

The only kill switch was `HIGHLIGHTS_LIVE` — a 13-line module constant that shipped as `true`, was never exported from the package barrel or the `exports` map (so a deep import was blocked by Node), and would have turned highlights off for *everyone* rather than per host. It was also read **twice, independently**: once for the fetch gate in the seam hook, once for row visibility in `bible-reader.tsx`. Two reads of one question that could in principle disagree.

**What is true now that this ships:** a host that says nothing about highlights gets a reader with zero highlight surface — no swatch row, no network, no storage, no dialogs, no redirect the reader can originate. A host that wants them says so in one prop. Success is measurable in the negative: on a default mount, DevTools Network shows no `/v1/highlights` request and no tap can navigate the user away.

- [packages/ui/src/components/bible-reader.tsx](https://github.com/youversion/platform-sdk-react/pull/301/files#diff-272a5534dafd6ceec55ef3874230e61cc594c7e0dd66e611400fbcafdc73aa3e) — new public prop `enableHighlights?: boolean` on `BibleReader.Root`, defaulting to `false`; plus two dev-only misconfiguration warnings
- [packages/ui/src/lib/feature-flags.ts](https://github.com/youversion/platform-sdk-react/pull/301/files#diff-a3cd317b440434732a0ea11fc26f11b75119531a02be59dbc00481addf0335d8) — **deleted**. `enableHighlights` is the sole gate
- [.changeset/bible-reader-highlights-opt-in.md](https://github.com/youversion/platform-sdk-react/pull/301/files#diff-12e10ef4f39c3c4b906f3b83adc52ea6f884e5c7c5aa7997183a33d80c8425eb) — minor bump across all three fixed-version packages (`2.5.0`), leading with the behavior change and the one-line migration
- [packages/ui/src/components/bible-reader.stories.tsx](https://github.com/youversion/platform-sdk-react/pull/301/files#diff-9bc69a287606531da07bf40451b0a8537116f15dadfbcf5f21fbe654e22987ba) — `SelfContained` (integration-tagged, with a `play` function) and `Self-contained (not opted in)`. The live self-contained path had **no** story or example coverage before this
- [CONTEXT.md](https://github.com/youversion/platform-sdk-react/pull/301/files#diff-86fee53c33308768d49a4ac0ff5720fd7f5411a8c8bd62ab9b80bbffa2977a14) — new **Highlights opt-in** glossary term, with an `_Avoid_` line separating it from the internal `highlightsEnabled` popover prop
- [examples/vite-react/src/pages/BibleReaderPage.tsx](https://github.com/youversion/platform-sdk-react/pull/301/files#diff-5408e8d168128bdbf8d8a7925f1580d1d66a5b26871bae3233b158eb511be37f) — the example opts in, with a comment naming the prop

**Not changed:** verse selection, Copy, and Share behave identically in every configuration. Controlled mode is untouched and ignores `enableHighlights` — passing `highlights` was always its own opt-in.

## How I implemented it

Three commits, one per phase, each a vertical slice.

### 1 · The gate swap, end to end — [`48e9f88`](https://github.com/youversion/platform-sdk-react/commit/48e9f88)

Public prop → context → `Content` → seam hook → machine input → rendered DOM. This had to be one commit: the flag cannot be deleted before its replacement exists, and the six suites calling `setHighlightsLive` break the moment it does.

- [bible-reader.tsx](https://github.com/youversion/platform-sdk-react/pull/301/files#diff-272a5534dafd6ceec55ef3874230e61cc594c7e0dd66e611400fbcafdc73aa3e) — `RootProps` gains the prop; `Root` destructures it as `enableHighlights = false`; the context value sets `enableHighlights: isHighlightsControlled ? false : enableHighlights`, so controlled mode can never light the self-contained path. Everything downstream leans on that one expression.
- [use-bible-reader-highlights.ts](https://github.com/youversion/platform-sdk-react/pull/301/files#diff-231a7c0be98b61c1066e7049e264eef7051544f7506bcbe70e0b54315f7fb0c9) — the opt-in becomes an input, and the composed gate becomes an output: `highlightsAvailable = optedIn && hasAuthProvider`. That is the **same expression** as the machine's `isEnabledNow` guard, so the color row and the statechart now provably agree instead of agreeing by coincidence. `bible-reader.tsx` consumes it rather than re-deriving.
- [bible-reader-highlights-machine.ts](https://github.com/youversion/platform-sdk-react/pull/301/files#diff-7881be8405a3a87d084a0d8d12cc8608d855f17bee5b2404983f28d7348aee64) — `flagOn` renamed `enableHighlights` across input, context, `AUTH_CHANGED`, and both enable guards. Pure rename; one vocabulary from public prop to statechart.
- [feature-flags.ts](https://github.com/youversion/platform-sdk-react/pull/301/files#diff-a3cd317b440434732a0ea11fc26f11b75119531a02be59dbc00481addf0335d8) — deleted.
- [use-bible-reader-highlights.test.tsx](https://github.com/youversion/platform-sdk-react/pull/301/files#diff-1be915834330087a37d778ad6a27b8307248d42c90b1e8e0f7ccdcab256d0ea2) plus five sibling suites — migrated from the `beforeEach`/`afterEach` flag dance to per-render input. `describe('… — flag off (dark launch)')` became `describe('… — not opted in (default)')`, now the most load-bearing assertion in the file because it tests what actually ships.
- [bible-reader.test.tsx](https://github.com/youversion/platform-sdk-react/pull/301/files#diff-2e4fb1903dceca9d60e0f8dbd60aaf14b5fa160cf6496d4d80244d6acf90dcf3) — new Root-level describe covering the public prop contract, which had no coverage at all.

### 2 · Dev-time misconfiguration warnings — [`e6efc8d`](https://github.com/youversion/platform-sdk-react/commit/e6efc8d)

Both follow the one-shot `!IS_PRODUCTION` shape already used by the controlled-mode latch warning. `IS_PRODUCTION` treats an environment with no `process` global as production, which is what makes this safe inside the React Native Expo DOM webview.

- **`Root`** warns when `onHighlightApply` / `onHighlightRemove` is passed without a `highlights` prop. Those callbacks only fire in controlled mode; in self-contained mode they were silently never called. The message names the prop actually passed and gives both exits.
- **`Content`** warns when `enableHighlights` is set with no auth provider mounted, naming `YouVersionProvider` + `includeAuth` + `authRedirectUrl`. Placed in `Content` rather than `Root` because the seam already computed the answer — putting it in `Root` would mean a second `useContext(YouVersionAuthContext)` read, the exact re-derivation this PR removes.

### 3 · Story, docs, glossary, changeset — [`5e1711b`](https://github.com/youversion/platform-sdk-react/commit/5e1711b)

- [handlers.ts](https://github.com/youversion/platform-sdk-react/pull/301/files#diff-9e74c20bb960ccdaaf69957081ed028e695e93a3b5359d814792ea31c6145e0f) — the global story MSW set had no `/v1/highlights` handler at all. Added GET/POST/DELETE.
- [bible-reader.stories.tsx](https://github.com/youversion/platform-sdk-react/pull/301/files#diff-9bc69a287606531da07bf40451b0a8537116f15dadfbcf5f21fbe654e22987ba) — the `play` function asserts fetched highlights paint on load, the swatch row renders *because the host opted in*, and an optimistic apply lands. It seeds the grant explicitly: `setupAuthenticatedUser()` seeds only the session, so without that a tap would land in the permission dialog instead of the write path.
- [YPE-3705 ADR](https://github.com/youversion/platform-sdk-react/pull/301/files#diff-066cab1e8dad08e0b6a7a2aba3b369c14b074b049f47088965e555c1a1ee5ff6) and [highlight-flow-statechart.md](https://github.com/youversion/platform-sdk-react/pull/301/files#diff-82773c380ba52dade68aed798bb8eedc7a9e4c75e21e107f8a6bf3e2b15721a8) — opt-in vocabulary throughout. The ADR's decisions stand as written; only the referent of "the gate" changed.

## Deviations from the plan

No plan file — this task ran research → design discussion → **structure outline** → phased implementation. Comparing against the [structure outline](https://cloud.humanlayer.com/artifacts/019fa46b-90ff-7741-914f-76803f03982b):

### Implemented as planned

- `enableHighlights` on `BibleReader.Root`, default `false`, ignored in controlled mode
- `feature-flags.ts` deleted; `highlightsAvailable` returned from the seam; the double read at `bible-reader.tsx:667` removed by construction
- All six suites migrated off `setHighlightsLive`; machine tests needed only a rename
- Both dev warnings, in the components the outline specified
- `SelfContained` story, MSW handlers, glossary/ADR/statechart updates, minor changeset

### Deviations/surprises

- **`highlightsInteractive` was deleted, not kept.** The outline had the seam returning both booleans. After the swap, `highlightsInteractive` had exactly one consumer — the Phase 2 auth warning — where `enableHighlights && !highlightsAvailable` is provably identical (with the opt-in true, `highlightsAvailable` reduces to `hasAuthProvider`). Every test asserted the two identically, which was the tell. The seam now exposes **one** enablement output. Its auth-axis rationale moved into `highlightsAvailable`'s JSDoc rather than being lost.
- **The hook-local is named `optedIn`, not `flagOn`.** Keeping `flagOn` alongside the renamed machine field would have produced `input: { enableHighlights: flagOn }` — the exact split vocabulary this change exists to remove.
- **The ADR still names `HIGHLIGHTS_LIVE`.** The outline asked for both a superseded-flag note *and* a grep gate returning zero hits — mutually exclusive. The gate exists to catch *stale* references describing the flag as live; an ADR is a historical record, and "see git history for the constant it exported" is worse for a future reader. The ADR names all three symbols; the gate carves out that one file.
- **The `Content` warning omits a `!isHighlightsControlled` clause.** The context value already forces `enableHighlights` to `false` in controlled mode, so the clause would be dead code. Worth knowing the two guards are coupled.

### Additions not in plan

- **Fixed a latent bug in two existing tests.** Two scope-change cases called `rerender({ versionId: 111, book: 'JHN', chapter: '4' })` without spreading `defaultOptions` — harmless while enablement came from a module flag, broken the instant it became a per-render input. Both assertions were weaker than they looked. Fixed by spreading.
- `bible-reader-controlled.test.tsx`'s "ignores onHighlightApply / onHighlightRemove in self-contained mode" is literally the misconfiguration the new warning targets, so it now asserts the warning fired instead of leaking it to stderr. Two comments in that file also claimed "HIGHLIGHTS_LIVE off" when the mechanism actually exercised was the absence of an auth provider — corrected.
- Two extra warning test cases beyond the outline's five, and a `Self-contained (not opted in)` story showing the default posture.
- `packages/ui/AGENTS.md` — the canonical `BibleReader` snippet gains the prop.

### Items planned but not implemented

None.

### Review fix — duplicate write on withdraw/restore (`8b71470`)

Greptile flagged a P1 on `bible-reader-highlights-machine.ts` and it reproduced. Because `enableHighlights` is deliberately **not latched**, a host can withdraw the opt-in mid-write. `resetWriteStateIfSignedOut` only fired on `!isAuthenticated`, so withdrawing while still signed in preserved `overlay`, `reconcile`, `writeIntent`, and `queue`.

Leaving `enabled` stops the `writer.writing` invoke, so the in-flight write can never settle — `settleWrite`/`shiftQueue` never run and the op stays at `queue[0]`. Re-entering `enabled` hit `writer.idle.always → queueHasWork` and **re-sent the request**: a duplicate POST whose first execution may already have reached the server, or a DELETE against an already-deleted highlight. The repro test asserted `createHighlight` called once and got two.

Fixed by broadening the action to `resetWriteStateIfInert`, which now fires on sign-out **or** on going disabled (`!enableHighlights || !hasAuthProvider` — the `isDisabledNow` condition). Dropping the queue is strictly safer than replaying it: the refetch that fires when the adapter's `live` flips back on brings server truth either way. The sign-out trigger is kept and still covered, since sign-out does *not* leave `enabled` (a signed-out tap must still reach the sign-in dialog), so nothing else would clear the overlay there.

The pending-highlight `sessionStorage` stash is deliberately **not** cleared — a queued write is work the host just cancelled, while a stashed intent is a user action awaiting a redirect round-trip with its own 10-minute TTL. That asymmetry is now documented in `docs/highlight-flow-statechart.md`.

Three tests added covering the in-flight write, the write queued behind it, and the sign-out path that must keep working.

## How to verify it

```bash
git fetch origin && git checkout make-bible-highlights-opt-in-feature
pnpm install && pnpm build
```

### Manual Testing

- [ ] `pnpm dev:web` with a real `VITE_YVP_APP_KEY` → select a verse: swatch row present, tap paints, reload persists
- [ ] Remove `enableHighlights` from `examples/vite-react/src/pages/BibleReaderPage.tsx` → swatch row gone, Copy/Share still present, **zero `/v1/highlights` requests in DevTools Network**, no tap can trigger a redirect
- [ ] Move `enableHighlights` onto a reader outside the auth-enabled provider → console warns exactly once (not once per render), no crash, reader renders normally
- [ ] `pnpm --filter @youversion/platform-react-ui storybook` → `SelfContained` paints on load and on tap; `Self-contained (not opted in)` shows Copy/Share only

### Automated Tests

```bash
pnpm typecheck
pnpm lint
pnpm test                                              # core 369 · hooks 289 · ui 371
pnpm --filter @youversion/platform-react-ui test:integration   # 37 files, 417 tests
pnpm build
```

### Release note for reviewers

**The release gate is cleared.** The React Native Expo SDK's `BibleReader` runs controlled mode — it passes a `highlights` prop, so it is unaffected by the default flip and nothing needs to land in that repo in this release window. A self-contained RN integration would have required coordinating both repos.

`BibleReader` highlights are now opt-in via a new `enableHighlights` prop on `BibleReader.Root` (default `false`); the internal `HIGHLIGHTS_LIVE` dark-launch flag is removed.

<h3>Greptile Summary</h3>

Updates `BibleReader` highlights to require explicit host opt-in.
- Adds the `enableHighlights` prop and threads it through the reader context, highlights adapter, and state machine.
- Keeps controlled highlights mode independent of the new self-contained-mode gate.
- Clears queued and optimistic write state when highlights become inert, preventing cancelled writes from replaying after re-enablement.
- Removes the internal highlights feature flag and updates tests, stories, examples, documentation, and release notes.

The PR appears safe to merge.

The previously reported queued-write replay path is addressed by clearing all write state before entering the disabled state, while the writer invocation is cancelled and fresh server state is fetched after re-enablement; no blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/ui/src/components/bible-reader.tsx | Adds the public opt-in prop, propagates the effective self-contained enablement value, and introduces development-only configuration warnings. |
| packages/ui/src/components/use-bible-reader-highlights.ts | Replaces the module feature flag with the per-reader opt-in and exposes one composed availability gate for fetching, machine input, and rendering. |
| packages/ui/src/components/bible-reader-highlights-machine.ts | Renames the enablement input and safely clears write state when the self-contained path becomes inert, resolving the previously reported replay condition. |
| packages/ui/src/components/bible-reader-highlights-machine.test.ts | Adds focused coverage confirming that in-flight and queued operations are not replayed after opt-in withdrawal and restoration. |
| packages/ui/src/components/bible-reader.stories.tsx | Adds opted-in and default-off self-contained stories with integration coverage for fetched and optimistically applied highlights. |
| packages/ui/src/test/mocks/handlers.ts | Adds the highlight API handlers required by the new self-contained Storybook integration path. |
| .changeset/bible-reader-highlights-opt-in.md | Documents the behavior change and applies the required unified minor release across all three packages. |

<h3>Flowchart</h3>

```mermaid
flowchart TD
  Root["BibleReader.Root"] --> Mode{"Controlled highlights?"}
  Mode -->|Yes| Controlled["Project host-supplied highlights"]
  Controlled --> Intent["Emit apply/remove intents"]
  Mode -->|No| OptIn{"enableHighlights?"}
  OptIn -->|No| Disabled["Disabled: no UI, fetches, writes, or dialogs"]
  OptIn -->|Yes| Provider{"Auth provider mounted?"}
  Provider -->|No| Disabled
  Provider -->|Yes| Machine["Self-contained highlights machine"]
  Machine --> Fetch["Fetch server highlights when authenticated"]
  Machine --> Flow["Handle consent and sign-in flow"]
  Machine --> Writer["Serialize optimistic writes"]
  Writer --> Withdraw{"Opt-in/provider withdrawn?"}
  Withdraw -->|Yes| Reset["Clear overlay, reconciliation, ownership, and queue"]
  Reset --> Disabled
```

<sub>Reviews (2): Last reviewed commit: ["fix(ui): drop write state when the highl..."](https://github.com/youversion/platform-sdk-react/commit/8b71470d9b02adcd22ad96178939bc1829f964b4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=47660740)</sub>

**Context used:**


jaredhightower-youversion (commented):
Nice work on this one. The opt-in latch, the misconfiguration warnings, and the write-state cleanup on withdrawal are all thoughtfully done, and the test coverage across the machine, hook, and component layers is thorough.

I've left a handful of non-blocking comments, mostly small dedup and trim opportunities in docs, tests, and mocks. Take or leave any of them.

packages/ui/src/components/bible-reader.tsx
```diff
@@ -545,6 +578,9 @@ function Root({
     onShare,
     highlights: isHighlightsControlled ? highlights : undefined,
     isHighlightsControlled,
+    // Controlled mode can never light the self-contained path, whatever the
+    // host passes here.
+    enableHighlights: isHighlightsControlled ? false : enableHighlights,
```
jaredhightower-youversion, line 583: **suggestion (non-blocking):** The controlled-mode normalization duplicates a guard the hook already owns.

I think the `isHighlightsControlled ? false : enableHighlights` normalization here is a second copy of a guard the hook already owns — `useBibleReaderHighlights` computes `!isControlled && enableHighlights` and documents that controlled mode ignores the flag. Could we pass `enableHighlights` through as-is and just add `!isHighlightsControlled &&` to the no-auth warning condition in `Content`? That keeps the warning quiet in controlled mode while dropping the ternary, its comment, and the "already resolved against the mode latch" doc on the context field.

packages/ui/src/test/mocks/handlers.ts
```diff
@@ -7,6 +7,69 @@ import mockLanguages from '../mock-data/languages.json';
...
+      ],
+    });
+  }),
+
+  http.post('*/v1/highlights', async ({ request }) => {
+    const body = (await request.json()) as {
+      request_id?: string;
+      highlight?: { bible_id?: number; passage_id?: string; color?: string };
+    };
+    if (
```
jaredhightower-youversion, line 42: **suggestion (non-blocking):** All three handlers carry dead 400 validation branches.

I think the 400 validation branches in all three handlers here can go — the only caller is the SDK itself, which always sends well-formed requests (core validates before sending and generates `request_id` itself), and no story or test asserts any of these 400s. The GET handler's missing-param check, this POST body-shape check (plus the structural type annotation that only feeds it), and the DELETE `bible_id` check (plus its `new URL()` parse) are all unreachable. The DELETE collapses to `http.delete('*/v1/highlights/:passageId', () => new HttpResponse(null, { status: 204 }))` and the POST to echoing `body.highlight` with a 201 — about 33 lines total.

packages/ui/src/components/bible-reader-highlights-machine.ts
```diff
@@ -393,15 +399,35 @@ export const bibleReaderHighlightsMachine = setup({
...
+        enableHighlights: event.enableHighlights,
         hasAuthProvider: event.hasAuthProvider,
         isAuthenticated: event.isAuthenticated,
       };
     }),
 
-    /** Sign-out / going disabled must not leave a stale overlay to resurface later. */
-    resetWriteStateIfSignedOut: assign(({ event }) => {
-      if (event.type !== 'AUTH_CHANGED' || event.isAuthenticated) return {};
+    /**
```
jaredhightower-youversion, line 408: **suggestion (non-blocking):** This 19-line JSDoc on a 6-line action duplicates the statechart doc.

I think this JSDoc could shrink to just the two triggers and the why — the whole writer-region replay walkthrough (writer.writing invoke, settleWrite/shiftQueue, queueHasWork, duplicate POST) is already in docs/highlight-flow-statechart.md almost word for word, and having it in two places means one will drift when the writer region changes. Something like: "Sign-out or a withdrawn opt-in must not leave stale write state: sign-out keeps the machine in `enabled` so nothing else clears the overlay, and going disabled strands an in-flight write at queue[0] that would replay on re-enable — dropping is safe since the refetch on re-enable restores server truth (see docs/highlight-flow-statechart.md for the mechanism)."

docs/highlight-flow-statechart.md
```diff
@@ -13,8 +15,26 @@ to explore it interactively.
 ## States and events
 
 - **`booting`** → routes to `disabled` or `enabled` from the initial input.
-- **`disabled`** — the flag is off **or** no auth provider is mounted. Fully
+- **`disabled`** — the host did not opt in (`enableHighlights` is false, the
+  default) **or** no auth provider is mounted. Fully
   inert: no fetch, no writes, no dialogs. A color tap resolves to `noop`.
+
+  Entering `disabled` **drops all write state** (`overlay`, `reconcile`,
```
jaredhightower-youversion, line 22: **suggestion (non-blocking):** The 16-line `disabled` rationale restates the `resetWriteStateIfInert` comment.

I think the `disabled` rationale block here restates the `resetWriteStateIfInert` doc comment in bible-reader-highlights-machine.ts almost sentence for sentence — could we collapse it to one line plus a pointer, the same way this doc already handles the vapor fix ("see the block comment... single source of truth")? One caveat: the pending-highlight-stash asymmetry note only lives here, not in the code comment, so keep that sentence (or move it into the action's comment before pointing at it).

(This and the machine JSDoc comment are two halves of one dedup — pick which side keeps the prose.)

packages/ui/src/components/bible-reader.test.tsx
```diff
@@ -376,3 +425,175 @@ describe('BibleReader Toolbar - onChapterPickerPress', () => {
     expect(screen.queryByPlaceholderText('Search')).not.toBeInTheDocument();
   });
 });
+
+/**
+ * The public opt-in contract on `BibleReader.Root`. Every case renders inside an
+ * auth provider, so the auth axis is held constant and `enableHighlights` is the
+ * only thing being measured.
+ */
+function AuthWrapper({ children }: { children: ReactNode }) {
```
jaredhightower-youversion, line 434: **suggestion (non-blocking):** Local `AuthWrapper` re-implements `Providers` from test utils.

I think the local `AuthWrapper` here can go — `@/test/highlights-test-utils` (which we already import `mockUserInfo` from) exports `Providers` with the exact same auth context value by default, and it's what the sibling highlights integration tests use. Swapping `<AuthWrapper>` for `<Providers>` at the six call sites would drop ~11 lines and keep both suites on one provider stack.

packages/ui/src/components/bible-reader.test.tsx
```diff
@@ -376,3 +425,175 @@ describe('BibleReader Toolbar - onChapterPickerPress', () => {
...
+  return (
+    <YouVersionAuthContext.Provider
+      value={{ userInfo: mockUserInfo, setUserInfo: vi.fn(), isLoading: false, error: null }}
+    >
+      {children}
+    </YouVersionAuthContext.Provider>
+  );
+}
+
+function readerJsx(props: Partial<BibleReaderRootProps> = {}) {
```
jaredhightower-youversion, line 444: **suggestion (non-blocking):** The new suites re-implement the harness from bible-reader-controlled.test.tsx.

`readerJsx` / `renderReader` / `selectVerse` / `getApplyButtons` here are near-verbatim copies of the same-named helpers in bible-reader-controlled.test.tsx. Worth extracting to a shared test util so the two DOM-poking harnesses don't drift independently?

packages/ui/src/components/bible-reader.test.tsx
```diff
@@ -376,3 +425,175 @@ describe('BibleReader Toolbar - onChapterPickerPress', () => {
...
+      { enabled: false },
+    );
+
+    selectVerse(container, 1);
+    // Copy / Share still open the popover — they never depended on highlights.
+    await waitFor(() => expect(screen.getByRole('dialog')).toBeTruthy());
+    expect(getApplyButtons()).toHaveLength(0);
+  });
+
+  it('renders no color row and disables the fetch when enableHighlights={false}', async () => {
```
jaredhightower-youversion, line 490: **suggestion (non-blocking):** The `enableHighlights={false}` test is a copy of the omitted-prop test.

I think these two opt-out tests could collapse into one `it.each([{}, { enableHighlights: false }])(...)` — they're identical except for the props, and spreading `{}` keeps the "prop omitted" case genuinely omitted. Saves ~12 lines with no coverage loss.

packages/ui/src/components/bible-reader-highlights-machine.test.ts
```diff
@@ -255,3 +261,107 @@ describe('bibleReaderHighlightsMachine — pending stash on lost permission', ()
...
+    expect(actor.getSnapshot().matches('enabled')).toBe(true);
+
+    // The stale op must not be re-sent. Server truth arrives via the refetch the
+    // adapter fires when `live` flips back on.
+    await Promise.resolve();
+    expect(createHighlight).toHaveBeenCalledTimes(1);
+    actor.stop();
+  });
+
+  it('clears queued (not yet started) writes when the opt-in is withdrawn', async () => {
```
jaredhightower-youversion, line 313: **suggestion (non-blocking):** This test exercises the same code path as the test above it.

I think the `clears queued (not yet started) writes` test can go — `resetWriteStateIfInert` drops the whole queue in one `assign`, so the 2-op case runs the exact same code as the in-flight test above it, which already checks `queue` is emptied and `createHighlight` isn't re-fired. If the "queued ops are dropped too" policy feels worth pinning, could we just add a second `TAP_COLOR` to that preceding test instead of keeping the full ~31-line duplicate?

packages/ui/src/components/use-bible-reader-highlights.test.tsx
```diff
@@ -110,6 +109,32 @@ describe('useBibleReaderHighlights — flag off (dark launch)', () => {
     expect(mocked.deleteHighlight).not.toHaveBeenCalled();
     expect(result.current.highlightedVerses).toEqual({});
   });
+
+  it('reports the color row as unavailable, even with an auth provider mounted', () => {
```
jaredhightower-youversion, line 113: **suggestion (non-blocking):** This test folds into the inert test above it.

I think this test could fold into the "is fully inert" one right above it — the render setup is identical (enableHighlights: false + AuthWrapper), so adding `expect(result.current.highlightsAvailable).toBe(false)` there covers it and saves a whole test block (maybe rename the merged test to mention the color row). The omitted-prop test below already keeps its own highlightsAvailable assertion, so no coverage is lost.

packages/ui/src/components/bible-reader.stories.tsx
```diff
@@ -810,6 +810,110 @@ export const JoshuaIntroChapter: Story = {
...
+  },
+};
+
+/**
+ * The shipped default: no `enableHighlights`, so the reader is inert on
+ * highlights even inside an auth-enabled provider with a signed-in user. The
+ * verse action bar offers Copy and Share only — no swatch row, no
+ * `/v1/highlights` traffic, no consent dialogs, no redirects.
+ */
+export const SelfContainedNotOptedIn: Story = {
```
jaredhightower-youversion, line 895: **suggestion (non-blocking):** `SelfContainedNotOptedIn` renders identically to existing stories.

I think we could drop the `SelfContainedNotOptedIn` story — it has no play function, and on load it renders identically to the opted-in stories since the missing swatch row only shows up after a verse tap that never happens. The not-opted-in behavior is already exercised more rigorously in bible-reader.test.tsx (both the omitted-prop and `enableHighlights={false}` cases tap a verse and assert no Apply buttons), and the default-off contract is documented in the `SelfContained` docblock.

packages/ui/src/components/bible-reader-highlights-machine.ts
```diff
@@ -393,15 +399,35 @@ export const bibleReaderHighlightsMachine = setup({
...
+     *    Re-entering `enabled` would hit `writer.idle.always → queueHasWork` and
+     *    re-send a request whose first execution may already have reached the
+     *    server — a duplicate POST, or a DELETE against an already-deleted
+     *    highlight. Dropping the queue is strictly safer than replaying it: the
+     *    refetch that fires when the adapter's `live` flips back on brings
+     *    server truth either way.
+     */
+    resetWriteStateIfInert: assign(({ event }) => {
+      if (event.type !== 'AUTH_CHANGED') return {};
+      const goingDisabled = !event.enableHighlights || !event.hasAuthProvider;
```
jaredhightower-youversion, line 429: **nitpick (non-blocking):** Double-negative guard.

The `goingDisabled` temp gets negated right back, so the guard reads as a double negative. I think the positive form is easier to scan — `if (event.isAuthenticated && event.enableHighlights && event.hasAuthProvider) return {};` — and the doc comment above already names the "going disabled" concept, so we don't lose anything by dropping the variable.

packages/ui/src/components/bible-reader-highlights-machine.test.ts
```diff
@@ -255,3 +261,107 @@ describe('bibleReaderHighlightsMachine — pending stash on lost permission', ()
...
+  /**
+   * `enableHighlights` is deliberately NOT latched, so a host can turn highlights
+   * off mid-session. Leaving `enabled` stops the `writer.writing` invoke, so an
+   * in-flight write can never settle — `settleWrite`/`shiftQueue` don't run and
+   * the op stays at `queue[0]`. Without clearing it, re-entering `enabled` hits
+   * `writer.idle.always → queueHasWork` and re-sends a request whose first
+   * execution may already have reached the server.
+   */
+  it('does not replay an in-flight write when the opt-in is withdrawn and restored', async () => {
+    const pending = deferred();
```
jaredhightower-youversion, line 275: **nitpick (non-blocking):** Deferred never resolves in these three new tests.

The whole point here is that the write never settles, so `vi.fn().mockReturnValue(new Promise(() => {}))` would make that intent explicit instead of implying a resolve that never comes. (The `deferred()` helper itself should stay — the earlier writeIntent tests genuinely resolve theirs.)

packages/ui/src/components/bible-reader-highlights-machine.test.ts
```diff
@@ -255,3 +261,107 @@ describe('bibleReaderHighlightsMachine — pending stash on lost permission', ()
     actor.stop();
   });
 });
+
+describe('bibleReaderHighlightsMachine — withdrawing the opt-in', () => {
+  /**
```
jaredhightower-youversion, line 266: **nitpick (non-blocking):** Describe-level JSDoc duplicates the `resetWriteStateIfInert` comment.

Same settleWrite/shiftQueue/queueHasWork story as the machine file, and two copies will drift. Could we shrink it to a one-liner like `// Withdrawal must drop in-flight/queued writes; see resetWriteStateIfInert.` and let the machine comment stay the single source of truth? The inline comment inside the test ("Still signed in, so the old sign-out-only reset would not have fired") carries the test-specific part and is worth keeping.

packages/ui/src/components/bible-reader-controlled.test.tsx
```diff
@@ -458,14 +458,20 @@ describe('BibleReader controlled mode - events', () => {
   });
 
   it('ignores onHighlightApply / onHighlightRemove in self-contained mode', async () => {
+    // Exactly the misconfiguration the Root dev warning names, so swallow it
+    // here and assert it fired rather than letting it leak into test output.
+    const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
     const onHighlightApply = vi.fn();
     const onHighlightRemove = vi.fn();
     const { container } = renderReader({ onHighlightApply, onHighlightRemove });
+    expect(warnSpy).toHaveBeenCalledWith(expect.stringContaining('will never be called'));
```
jaredhightower-youversion, line 467: **nitpick (non-blocking):** Warning-message assertion re-checks coverage that lives elsewhere.

I think the `expect(warnSpy).toHaveBeenCalledWith(expect.stringContaining('will never be called'))` here is re-asserting something bible-reader.test.tsx already covers in depth (the two ORPHANED_CALLBACK tests check the message, the callback name, and one-shot behavior). Since this test is really about the callbacks never firing, could we keep the spy purely as noise suppression with a one-liner like `// Silence the Root dev warning for this misconfiguration (asserted in bible-reader.test.tsx).`?

packages/ui/src/components/use-bible-reader-highlights.test.tsx
```diff
@@ -468,7 +493,7 @@ describe('useBibleReaderHighlights — scope changes', () => {
     });
 
     const { result, rerender } = renderHook(
-      (props: { versionId: number; book: string; chapter: string }) =>
+      (props: { versionId: number; book: string; chapter: string; enableHighlights: boolean }) =>
```
jaredhightower-youversion, line 496: **nitpick (non-blocking):** Hand-written prop type duplicates what TS infers.

This prop type is restated at three renderHook sites even though it's exactly what TS infers from `defaultOptions`. Could we use `(props: typeof defaultOptions) => useBibleReaderHighlights(props)` instead — or even point-free `renderHook(useBibleReaderHighlights, { wrapper: AuthWrapper, initialProps: defaultOptions })`? They'd stay in sync automatically if the options shape changes.

packages/ui/src/components/bible-reader.stories.tsx
```diff
@@ -810,6 +810,110 @@ export const JoshuaIntroChapter: Story = {
...
+        <BibleReader.Content />
+        <BibleReader.Toolbar />
+      </BibleReader.Root>
+    </div>
+  ),
+  play: async ({ canvasElement }) => {
+    const versesFor = (verse: number) =>
+      canvasElement.querySelectorAll<HTMLElement>(`.yv-v[v="${verse}"]`);
+
+    await waitFor(
```
jaredhightower-youversion, line 852: **nitpick (non-blocking):** First `waitFor` is subsumed by the paint check below it.

The highlight-paint `waitFor` right below can only pass after the whole chapter HTML has mounted (the fill is applied in a layout effect keyed on the html), so verse 3 is guaranteed to exist by then. Dropping the `versesFor(3).length > 0` block also brings this story in line with the others, which use a single waitFor for chapter load.

packages/ui/src/components/bible-reader.stories.tsx
```diff
@@ -810,6 +810,110 @@ export const JoshuaIntroChapter: Story = {
...
+      },
+      { timeout: 5000 },
+    );
+
+    // Selecting a verse opens the action bar — with the swatch row, which only
+    // renders because the host opted in.
+    await userEvent.click(versesFor(3)[0]!);
+
+    const swatchRow = await screen.findByRole('group', { name: /highlight colors/i });
+    await expect(swatchRow).toBeInTheDocument();
```
jaredhightower-youversion, line 872: **nitpick (non-blocking):** `toBeInTheDocument` after `findByRole` is redundant.

`findByRole` already throws if the swatch row never appears, so the query itself is the assertion. Could we drop this line (and the now-unused `swatchRow` binding) and just keep `await screen.findByRole('group', { name: /highlight colors/i })`? (Same pattern pre-exists elsewhere in this file, so scoping this to the new lines only.)

packages/ui/src/components/use-bible-reader-highlights.ts
```diff
@@ -392,12 +409,11 @@ export function useBibleReaderHighlights({
 
   return {
     highlightedVerses,
-    // Interactivity mirrors the machine's enabled/disabled gate: with no auth
-    // provider the machine is inert and the color row must not render. The flag
-    // is ANDed in by the caller. (`live` also folds in `isAuthenticated`, which
-    // we intentionally exclude here — a signed-out tap still enters the sign-in
-    // flow, so the row stays interactive.)
-    highlightsInteractive: hasAuthProvider,
+    // The one enablement output, identical to the machine's `isEnabledNow`. The
```
jaredhightower-youversion, line 412: **nitpick (non-blocking):** Comment restates the field's JSDoc.

This block restates the same two points (single enablement output, deliberate exclusion of `isAuthenticated`) that the JSDoc on the `highlightsAvailable` field already covers, and that JSDoc is what callers actually see on hover. The only unique bit — that the local `live` value also folds in isAuthenticated — could move into that JSDoc as one clause.

CONTEXT.md
```diff
@@ -40,10 +40,22 @@ the verse action popover. Never persisted; cleared on navigation.
...
+and the fetch additionally on an authenticated user — without the opt-in,
+or with no auth provider, the reader is inert: no swatch row, no fetches,
+no writes, no consent dialogs, nothing rendered from the API.
+
+## Highlights opt-in
+
+The host's explicit request for self-contained highlights: the
+`enableHighlights` prop on `BibleReader.Root`, default `false`. It is the
+single gate on the self-contained path — the SDK holds no second switch of
+its own (it replaced an internal dark-launch flag, now deleted). Ignored in
```
jaredhightower-youversion, line 54: **nitpick (non-blocking):** Historical parenthetical duplicated below.

I think the "(it replaced an internal dark-launch flag, now deleted)" parenthetical can go — the _Avoid_ line right below already marks the dark-launch flag as gone, and the full migration story lives in the YPE-3705 ADR. Ending the sentence at "no second switch of its own" keeps the glossary entry purely definitional.

docs/adr/YPE-3705-controlled-mode.md
```diff
@@ -74,12 +76,27 @@ path) via a `controlled` input, latched by `Root` at first mount:
...
-  the controlled prop surface ships while self-contained remains dark.
+- Controlled mode bypasses `enableHighlights` — the opt-in gates only the
+  self-contained server path, so the color row stays interactive for a
+  controlled host regardless of it.
 - The glossary (`CONTEXT.md`) defines "Controlled mode" and "Highlight
   intent"; intent and fact are deliberately never given the same name.
 
+## Superseded: the dark-launch flag (2026-07)
+
+The dark-launch flag module `packages/ui/src/lib/feature-flags.ts` — exporting
```
jaredhightower-youversion, line 87: **nitpick (non-blocking):** The flag → `enableHighlights` story is told three times in this ADR.

The bullet in "Non-obvious decisions", this Superseded section, and the Integration section all retell the dark-launch-flag history. Could this section shrink to about two lines — "## Superseded (2026-07) — `packages/ui/src/lib/feature-flags.ts` is deleted; the self-contained path is gated on `BibleReader.Root`'s `enableHighlights` prop (default `false`). Only the referent of 'the gate' changed." — and the line-43 bullet trim to just the decision ("That gate — now the host's `enableHighlights` opt-in — strictly means 'is the self-contained server path live'")? That keeps everything the "see the note below" reference needs.

examples/vite-react/src/pages/BibleReaderPage.tsx
```diff
@@ -3,7 +3,18 @@ import { BibleReader } from '@youversion/platform-react-ui';
 export function BibleReaderPage() {
   return (
     <div className="h-[calc(100vh-3.5rem)]">
-      <BibleReader.Root defaultBook="JHN" defaultChapter="1" defaultVersionId={3034}>
+      {/*
```
jaredhightower-youversion, line 6: **nitpick (non-blocking):** Comment restates the `enableHighlights` JSDoc.

I think we can drop most of this comment block — it restates the prop's JSDoc almost word-for-word, and duplicated prose in an example tends to drift from the prop docs. The one example-specific bit worth keeping is a one-liner like `{/* Requires the auth-enabled provider in ThemedApp.tsx */}`.

Dustin-Kelley:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

cameronapak:
[David has shared](https://teams.cloud.microsoft/l/message/19:bc453baaaf9c4aedb59018ada344ab83@thread.tacv2/1785275580596?tenantId=8c9a6ca9-b431-4964-afb4-b8e1a2ba636f&groupId=de52c8a3-9ed3-4869-80ac-dc210bc9bba2&parentMessageId=1785178900787&teamName=YouVersion&channelName=YV%20-%20Platform%20Team&createdTime=1785275580596):

> Thanks for the thoughts. Requiring developers need to turn on features one by one seems awkward in the long term. Once we add a few more features to BibleReader it would become really weird to tell people that to get the full set of features they'd need to turn them all on individually like `<BibleReader withHighlights withCustomColorHighlights withRelatedContent withNotes withAudio withSearch />`
>  
> I don't want that kind of complexity and messiness for SDK users. BibleReader is our reader experience; if they don't want its features, we're also giving them all the code and capabilities to build a custom reader however they want. Let's keep things simple.

Therefore, I'm closing this pull request

PR #304, opened 2026-07-28
feat(ui): swap the serif font from Source Serif 4 to Untitled Serif (YPE-1350, YPE-1910)

cameronapak:
> [!WARNING]
> - [x] This PR is awaiting #306 to first be merged in

[YPE-1350](https://lifechurch.atlassian.net/browse/YPE-1350) | [YPE-1910](https://lifechurch.atlassian.net/browse/YPE-1910)

The SDK's serif face was **Source Serif 4**, a Google Fonts stand-in adopted when [ADR-0001](docs/adr/0001-revert-brand-fonts-pending-licensing.md) pulled the hardcoded brand `@font-face` blocks out of the SDK. The BibleReader has therefore never rendered in YouVersion's brand serif.

ADR-0001's finding was about **delivery**, not the typeface: a hardcoded `@font-face` pointing at a CDN woff2 puts the file in front of third-party developer apps with nothing in the SDK able to gate it. So it set one condition for bringing a brand font back — load it via `/v1/fonts/{font_id}/stylesheet`, which requires an app key. Untitled Serif now does exactly that.

Shipped, this means: the BibleReader and every other serif surface render in Untitled Serif, delivered from the Fonts API using the app key consumers already supply — no font file in the repo, no new prop, no setup. A consumer who blocks the new hosts falls back to Source Serif 4 with no layout break, so the downside case is exactly today's behavior.

The sans stack is untouched and stays `'Inter', sans-serif`. This is a serif-only change.

- [packages/core/src/styles/theme.css](https://github.com/youversion/platform-sdk-react/pull/304/files#diff-7abe2a313c4913e40f49304f17861391884e584d219a89e897ecaacc1077c375) — `--yv-font-serif` becomes `'Untitled Serif', 'Source Serif 4', serif`, so **every** serif surface follows at once: BibleReader body text, the Bible card, `BibleText`, the version-picker abbreviation tile, footnotes, chapter headings, and the `lg` Verse of the Day card (this is YPE-1910's scope, done in the same pass)
- [packages/ui/src/components/YouVersionProvider.tsx](https://github.com/youversion/platform-sdk-react/pull/304/files#diff-b9f8fe57c37ebf08bbbf1b797f528191b0f06467398741565dd8444568fad3cdR61) — **new outbound request.** The provider now renders a hoisted `<link rel="stylesheet">` to `https://api.youversion.com/v1/fonts/1/stylesheet`, plus woff2 fetches from `cdn.youversion.com`. No new prop and no opt-out
- [packages/ui/src/components/bible-reader.tsx](https://github.com/youversion/platform-sdk-react/pull/304/files#diff-272a5534dafd6ceec55ef3874230e61cc594c7e0dd66e611400fbcafdc73aa3eR366) — the reader's **default font changes** to Untitled Serif, and readers whose saved preference is the old Source Serif stack are migrated on load so the picker still shows serif as active
- [packages/ui/src/i18n/locales/en.json](https://github.com/youversion/platform-sdk-react/pull/304/files#diff-68771d3e66a54763512c25b4a0c080d81102a40414f6c73ab77db6a0937a0531) — the font picker button reads **"Untitled"** instead of "Source Serif" (same in [es](https://github.com/youversion/platform-sdk-react/pull/304/files#diff-4df6549c1860e4369b8c7851352b7c33c8a475a5bb41caa5127675273997e113) / [fr](https://github.com/youversion/platform-sdk-react/pull/304/files#diff-408258fe86058d16edeb264f6d487b07e7d66b818685fa5711a46576fe8b250a) — a proper noun, so it isn't translated). **These three edits are the blocking CI failure** — see Open item
- [packages/ui/README.md](https://github.com/youversion/platform-sdk-react/pull/304/files#diff-0c5cfd3909673f57251cc7beaeb52268699df6f59bb230a52a9249b9bdf7115b) — **strict-CSP consumers must allowlist** `https://api.youversion.com` in `style-src` and `https://cdn.youversion.com` in `font-src`. This also removes a stale `storage.googleapis.com` entry for a host the SDK no longer calls
- [.changeset/untitled-serif-via-fonts-api.md](https://github.com/youversion/platform-sdk-react/pull/304/files#diff-91a74529dc77592f9833fc9a7037346f1eadb52a63cb98728e79874e14194a05) — minor across all three packages (new request + default-font change, so not a patch)

## How I implemented it

Five commits. The [walkthrough artifact](https://cloud.humanlayer.com/artifacts/019faac7-bc33-7d5b-a343-5548cc035b6c) has the same journey with inline diffs.

### 1 · Load the font, move the tokens (`7607cb5`)

- [packages/ui/src/lib/yv-fonts.tsx](https://github.com/youversion/platform-sdk-react/pull/304/files#diff-95647b903b4cdd844be56cc7d0d4461f011006c490b488029907ab627d15d9fb) *(new)* — builds `https://{apiHost}/v1/fonts/1/stylesheet?app_key={key}` and renders `<link rel="stylesheet" precedence="yv-sdk-fonts">`. This is the SDK's **first runtime-value-dependent stylesheet**: the URL needs the app key, which is only known at render time, so it can't live in the build-time-frozen `__YV_STYLES__` blob. React 19 hoists the link into `<head>` and dedupes by `href`, so multiple providers still yield one link. Returns `null` on a missing/whitespace key — no key, nothing to 401
- [YouVersionProvider.tsx](https://github.com/youversion/platform-sdk-react/pull/304/files#diff-b9f8fe57c37ebf08bbbf1b797f528191b0f06467398741565dd8444568fad3cdR61) — mounted after the `MissingAppKey` guard, so only the has-a-key branch requests it
- [theme.css](https://github.com/youversion/platform-sdk-react/pull/304/files#diff-7abe2a313c4913e40f49304f17861391884e584d219a89e897ecaacc1077c375) + [global.css](https://github.com/youversion/platform-sdk-react/pull/304/files#diff-beae923ec8489ad5d9fe003002ad65f5aae49ddbba3f4ff251f0f86e6fe03507) — the serif stack is declared **twice** and cannot be an alias: core can't import Tailwind, and `@theme inline` values are inlined into utilities rather than emitted as runtime custom properties. Both got the same literal, both got comments saying so
- [font-tokens.test.ts](https://github.com/youversion/platform-sdk-react/pull/304/files#diff-2e69f77f49661b74ba36fb132ccc0fe7c135ccaebdb2d68c6b0bf1d1c5fbcb63) *(new)* — reads both CSS files off disk and asserts the two declarations stay byte-identical, that Untitled Serif is first, and that Source Serif 4 is still in the stack *and* still `@import`-ed. Nothing else catches drift; jsdom never loads either sheet
- [yv-fonts.test.tsx](https://github.com/youversion/platform-sdk-react/pull/304/files#diff-3a8e6ef64eac4783dd94fa238d51e048e62383667665d52966fe05ec083c1876) *(new)* — URL shape, percent-encoding, custom `apiHost`, missing key, dedupe across two roots. The notable assertion is that **children still commit**: React can suspend a commit on `<link precedence>` until the sheet loads and jsdom never fires `load`, which would have meant mounting the provider could blank a consumer's tree

### 2 · Move the reader (`03f6aca`)

The token swap alone can't reach reader body text. `bible-reader.tsx` writes `--yv-reader-font-family` inline from its own state, and `bible-reader.css` applies it to every descendant — since that inline property is always defined, it wins over `--yv-font-serif`. So the reader needed its own changes:

- [verse-html-utils.ts](https://github.com/youversion/platform-sdk-react/pull/304/files#diff-b421c2c342ea4393a25e7c3e2918ccd70f0ec013214ccac701076588de8059e8) — adds `UNTITLED_SERIF_FONT`; `SOURCE_SERIF_FONT` is `@deprecated` but **kept**, because the migration needs the literal to compare against
- [bible-reader.tsx](https://github.com/youversion/platform-sdk-react/pull/304/files#diff-272a5534dafd6ceec55ef3874230e61cc594c7e0dd66e611400fbcafdc73aa3eR478) — the hydrate step maps exactly one legacy value forward: `savedFontFamily === SOURCE_SERIF_FONT ? UNTITLED_SERIF_FONT : savedFontFamily`. Deliberately **not** validation — `FontFamily` is `(string & {})` on purpose, so a host-supplied `defaultFontFamily="Georgia"` must still round-trip. Without the migration, a reader who'd already picked serif would reload to a picker with neither button active
- [bible-card.tsx](https://github.com/youversion/platform-sdk-react/pull/304/files#diff-506f3c33aef2f8facad99741d19d40d9816ee354ed2662981bf2e61449f67f03) — passes `fontFamily` explicitly rather than inheriting the token, so it needed the same swap
- [bible-reader.test.tsx](https://github.com/youversion/platform-sdk-react/pull/304/files#diff-2e4fb1903dceca9d60e0f8dbd60aaf14b5fa160cf6496d4d80244d6acf90dcf3) — two new hydrate tests (migrate the legacy value; leave a non-legacy value alone)
- [bible-reader.stories.tsx](https://github.com/youversion/platform-sdk-react/pull/304/files#diff-9bc69a287606531da07bf40451b0a8537116f15dadfbcf5f21fbe654e22987ba) — `LoadsSavedPreferencesFromLocalStorage` still seeds the *legacy* value, then asserts the migrated result on the CSS custom property, in localStorage, and on the active picker button

### 3 · Record the decision (`65c8815`)

- [docs/adr/0003-adopt-untitled-serif-via-fonts-api.md](https://github.com/youversion/platform-sdk-react/pull/304/files#diff-d53942f8d5d97ad8593d0ba762a6bc26bab37a096b8667a79c990d511d4fb5fa) *(new)* — records the delivery mechanism, the duplicated token, the reader migration, and the no-opt-out call
- [docs/adr/0001-…](https://github.com/youversion/platform-sdk-react/pull/304/files#diff-feccec964d92e74bbc53352376360816c36f44912f231a477f6b1495ce5c6000) — status becomes "Accepted, superseded **in part**", scoped to Untitled Serif

### 4 · Review passes (`f6a53a2`, `c7e3ebb`)

- **Public-repo hygiene.** Both ADRs publish with this repo, so the licensing narrative came out of both — including ADR-0003's CDN probe table, which printed an unauthenticated font URL and a compliance argument that don't belong in a public repo
- **Removed the other brand font entirely** ([ADR-0001](https://github.com/youversion/platform-sdk-react/pull/304/files#diff-feccec964d92e74bbc53352376360816c36f44912f231a477f6b1495ce5c6000), [ADR-0003](https://github.com/youversion/platform-sdk-react/pull/304/files#diff-d53942f8d5d97ad8593d0ba762a6bc26bab37a096b8667a79c990d511d4fb5fa), [theme.css](https://github.com/youversion/platform-sdk-react/pull/304/files#diff-7abe2a313c4913e40f49304f17861391884e584d219a89e897ecaacc1077c375), [global.css](https://github.com/youversion/platform-sdk-react/pull/304/files#diff-beae923ec8489ad5d9fe003002ad65f5aae49ddbba3f4ff251f0f86e6fe03507), [font-tokens.test.ts](https://github.com/youversion/platform-sdk-react/pull/304/files#diff-2e69f77f49661b74ba36fb132ccc0fe7c135ccaebdb2d68c6b0bf1d1c5fbcb63)). It documented a decision about a font the SDK doesn't use. **Verified it never shipped:** changeset `2ba9e6d` added it and `0d184fc` reverted it, both landing in `2.2.0`, so the released package never carried it. Anything implying Untitled Serif is not cleared came out too. Two [bible-chapter-picker.test.tsx](https://github.com/youversion/platform-sdk-react/pull/304/files#diff-f744764a5ece99e81b1b80128b3a6e82dbfca5919e98a35619775a188f1eaa03) test names still referenced it while asserting only size and weight — renamed to what they check
- **Scoped the suspense claim.** The comments in `yv-fonts.tsx` / `yv-fonts.test.tsx` asserted React "only suspends for stylesheets inserted during a transition." [React's docs](https://react.dev/reference/react-dom/components/link) say the component rendering a `precedence` stylesheet suspends while it loads, with no transition caveat. The test proves the synchronous path only, so it now says that and flags the transition path as uncovered — a latency risk on a transition-driven mount, not a hang, since failures settle too
- **Covered `YvFonts`' own guard.** The missing-key test routes through the provider's `MissingAppKey` branch, so `YvFonts`' `!appKey?.trim()` guard never executed. Now tested directly for undefined / empty / whitespace, and the leaked `console.error` spy is restored
- **README:** dropped the `font-display: swap` claim. That property comes from a stylesheet the Fonts API authors, not this SDK

Verified against the endpoint directly: `GET /v1/fonts/1/stylesheet` returns `401 Failed to resolve API Key` with no key and `401 Invalid ApiKey` with `?app_key=bogus` — so the query-param form the SDK builds is accepted by the gateway.

## Deviations from the plan

Plan: [04-structure-outline-untitled-serif-swap.md](.humanlayer/tasks/ype-1350-react-sdk-biblereader-use-correct-serif-font-swap-source-serif-for-untitled/04-structure-outline-untitled-serif-swap.md)

### Implemented as planned

- `<YvFonts />` matches the plan's snippet exactly — hardcoded font id 1, `apiHost` default, `precedence="yv-sdk-fonts"`, `null` on empty key
- Both token declarations, the drift-guard test, the reader default + migration + picker label, the i18n key rename, ADR-0003 and the ADR-0001 status update, the README CSP section, and the changeset

### Deviations/surprises

- **Reader body scope moved from Phase 1 to Phase 2.** The outline originally expected the token swap to change reader body text, then corrected itself in-document once the inline `--yv-reader-font-family` pin was found. The code follows the *corrected* plan. Not an implementation deviation — a plan self-correction
- **The migration unit test doesn't assert `--yv-reader-font-family`.** Reaching that property requires rendering `<BibleReader.Content />`, which calls `usePassage` and throws `YouVersion context not found` — that test file mocks the hooks package but renders no provider. The custom-property assertion lives in the story instead. Pre-flagged in the plan
- **ADR-0003 shipped substantially shorter than drafted.** The plan called for an ADR recording the full rationale; review established that this repo is public, so the licensing narrative and a CDN probe table were cut. What remains is what the code depends on
- **The i18n step is now known to be wrong as planned.** The outline said to rename the key in all three locale files directly. Repo policy says those files are upstream-owned — see Open item

### Additions not in plan

- `font-tokens.test.ts` also guards the **sans** token pair for drift, not just serif — same pattern, no reason to leave the other half unguarded
- The migration story additionally asserts the migrated value was **re-persisted** to localStorage, beyond the plan's "assert the migrated one on the CSS property"
- The `c7e3ebb` docs cleanup was not planned work; it came out of review

### Items planned but not implemented

None.

## Open item — locale ownership

**This is the one red check.** `.github/scripts/check-locale-ownership.sh` fails any PR touching `packages/ui/src/i18n/locales/**` unless it's a `platform-localization` sync PR. This PR renames `sourceSerifFontName` → `untitledSerifFontName` in all three.

It's a chicken-and-egg, and **both halves were reproduced locally**: revert the locale files and `i18n Check` fails instead, because `bible-reader.tsx` calls `t('untitledSerifFontName')` and the key wouldn't exist — that check hard-fails on a `t()` reference missing from `en.json`.

The only clean path is the documented one in [`docs/i18n-guidelines.md`](docs/i18n-guidelines.md):

1. Add `react.untitledSerifFontName` upstream in `platform-localization` (`sources/common/en.json`)
2. Merge there → `distribute-react.yml` opens a `chore/localization-sync-react-*` PR here
3. Rebase this PR and drop the three locale edits

Until that lands, this PR can't go green either way. Removing `react.sourceSerifFontName` upstream should be a **follow-up after this merges**, not part of step 1 — removing it while `main` still calls `t('sourceSerifFontName')` would turn `i18n Check` red on `main`.

## Not changed on purpose

- **`packages/{ui,core,hooks}/CHANGELOG.md`** still carry the original 2.2.0 entries, including the font naming this PR cleaned out of the ADRs. Rewriting shipped release notes falsifies the record
- **ADR-0001 keeps its `…-pending-licensing.md` filename**, even though the title no longer says that. Three published CHANGELOG entries link to that path and renaming would break them. Happy to rename if you'd rather take the dead links

## How to verify it

```bash
git fetch origin ype-1350-react-sdk-biblereader-use-correct-serif-font-swap-source-serif-for-untitled
git checkout ype-1350-react-sdk-biblereader-use-correct-serif-font-swap-source-serif-for-untitled
pnpm install && pnpm build
```

### Manual Testing

- [ ] `pnpm dev:web` with a real app key → BibleReader body text renders **Untitled Serif** (not Source Serif 4). DevTools → Network shows one request to `api.youversion.com/v1/fonts/1/stylesheet` and woff2 fetches from `cdn.youversion.com`
- [ ] Open reader settings → the serif button reads **"Untitled"** and is active by default
- [ ] Seed the legacy preference, reload, confirm it migrates:
      `localStorage.setItem('youversion-platform:reader:font-family', '"Source Serif 4", serif')` → after reload the value is the Untitled stack and the picker still shows serif active
- [ ] Pass a custom `defaultFontFamily="Georgia"`, pick it, reload → still Georgia (migration must not clobber host values)
- [ ] Block `api.youversion.com` (DevTools request blocking or a strict CSP) → serif text falls back to **Source Serif 4** with no layout break
- [ ] Bible card, `BibleText`, footnotes, chapter headings, and the `lg` VOTD card all render Untitled Serif (YPE-1910 scope)
- [ ] Apply the README's CSP block verbatim in a scratch host app and confirm nothing is blocked

### Automated Tests

```bash
pnpm test        # 30 files / 370 unit tests
pnpm typecheck
pnpm lint
pnpm build
```

Swap the SDK's serif face from Source Serif 4 to Untitled Serif, YouVersion's brand serif, loaded from the YouVersion Fonts API — strict-CSP consumers must allowlist `api.youversion.com` (`style-src`) and `cdn.youversion.com` (`font-src`).

[YPE-1350]: https://lifechurch.atlassian.net/browse/YPE-1350?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[YPE-1910]: https://lifechurch.atlassian.net/browse/YPE-1910?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

<h3>Greptile Summary</h3>

The PR adopts Untitled Serif as the SDK-wide serif face.
- Loads the authenticated Fonts API stylesheet from `YouVersionProvider` within a local Suspense boundary.
- Updates core and UI font tokens while retaining Source Serif 4 as the fallback.
- Changes the BibleReader and BibleCard defaults and migrates the legacy saved reader preference.
- Adds font-loading, token-consistency, and preference-migration coverage plus CSP and ADR documentation.

The PR appears safe to merge.

The previously reported locale-ownership failure is resolved at current HEAD: this PR no longer changes generated locale catalogs, and the upstream-synced English key satisfies the reader’s translation lookup.

| Filename | Overview |
|----------|----------|
| packages/ui/src/components/YouVersionProvider.tsx | Mounts the new font stylesheet loader behind the existing app-key guard and scopes stylesheet suspension locally. |
| packages/ui/src/lib/yv-fonts.tsx | Constructs the encoded Fonts API stylesheet URL and relies on React stylesheet hoisting and deduplication. |
| packages/ui/src/components/bible-reader.tsx | Changes the reader default and picker to Untitled Serif while migrating the exact legacy Source Serif preference. |
| packages/ui/src/styles/font-tokens.test.ts | Guards consistency between the duplicated core and UI font stacks and verifies the fallback remains loaded. |
| packages/core/src/styles/theme.css | Promotes Untitled Serif to the first SDK-wide serif choice while retaining Source Serif 4 and generic serif fallbacks. |
| packages/ui/src/styles/global.css | Mirrors the updated serif stack in the Tailwind theme consumed by UI utilities. |

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
  participant App as Consumer App
  participant Provider as YouVersionProvider
  participant Fonts as Fonts API
  participant CDN as Font CDN
  Provider->>Fonts: "GET /v1/fonts/1/stylesheet?app_key=..."
  Fonts-->>Provider: Untitled Serif stylesheet
  Provider->>CDN: Fetch referenced woff2 files
  CDN-->>Provider: Untitled Serif resources
  Provider-->>App: Render with Untitled Serif or Source Serif 4 fallback
```

<sub>Reviews (9): Last reviewed commit: ["Merge branch &#39;main&#39; into ype-1350-react-..."](https://github.com/youversion/platform-sdk-react/commit/6b7ed927a3ef494a86ad0d3383e33a7a7c6e53a6) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48128447)</sub>

**Context used:**


docs/adr/0003-adopt-untitled-serif-via-fonts-api.md
```diff
@@ -0,0 +1,186 @@
...
+defensible reading of "security measures so only authorized users can get the font" —
+but it is a reading, not a technical guarantee, and this ADR is the place that says so.
+
+**Known gap in the discovery half of the argument, outside this repo's control:** the
+public docs page at [developers.youversion.com/api/fonts](https://developers.youversion.com/api/fonts)
+publishes the 400-normal CDN URL verbatim in its example response, readable with no
+auth. That undercuts "you cannot learn the URL without a key," and rotating the CDN
+path would also require updating that example. Routed to whoever owns the API docs; not
+a blocker for this change and not fixable from this codebase.
+
```
cameronapak: issue: Important info that doesn't necessarily need to be public-facing. Should be a Jira task.

docs/adr/0001-revert-brand-fonts-pending-licensing.md
```diff
@@ -4,37 +4,33 @@ Date: 2026-06-24
...
-  agencies, partners, vendors, contractors, freelancers). Whether a Platform
-  developer is a "partner" is an **open legal question**.
-
-A "browser-consumable stylesheet" endpoint exists
-(`GET /v1/fonts/{font_id}/stylesheet`, accepts `app_key`, gateway injects the
-app-id header). It is the correct future consumption pattern, but it does **not**
-by itself resolve licensing: the woff2 it references still sits at a public CDN
-URL, so switching to it does not make the font file un-downloadable.
+- **Aktiv Grotesk App** as the sans default (`--yv-font-sans`), loaded via a hardcoded
+  `@font-face` pointing at a CDN woff2.
```
cameronapak: issue: Remove this entirely. We do not have permission to use Aktiv Grotesk font in our web SDK at this time. We use "Inter" in leiu of it.

docs/adr/0001-revert-brand-fonts-pending-licensing.md
```diff
@@ -4,37 +4,33 @@ Date: 2026-06-24
...
+- **Untitled Serif** as the serif default (`--yv-font-serif`) and the Bible Version
+  picker abbreviation tile, same hardcoded `@font-face` pattern.
+
+Neither font had licensing cleared for that delivery pattern at the time. The SDK's
+purpose is to render inside **third-party developer apps**, so a hardcoded `@font-face`
+puts the font file in front of third parties with nothing in the SDK able to gate it —
+access control is enforced server-side (API gateway + CDN), not in the SDK.
```
cameronapak: [issue](https://conventionalcomments.org/): Please remove the oversharing that's not needed. Thank you

docs/adr/0001-revert-brand-fonts-pending-licensing.md
```diff
@@ -44,24 +40,22 @@ Revert **both** brand fonts to the prior fallbacks for the shipping PR:
...
 
 Remove both brand `@font-face` blocks, the `--font-aktiv` / `--font-untitled-serif`
-aliases, the `yv:font-aktiv` / `yv:font-untitled-serif` usages, and the brand
-options in the Bible Reader font picker. The abbreviation-tile redesign and all
-other Figma layout/typography work, the `useOrganizations` hooks, and publisher
-names are retained — only the font **family** is reverted.
+aliases, the `yv:font-aktiv` / `yv:font-untitled-serif` usages, and the brand options in
+the Bible Reader font picker. The abbreviation-tile redesign and all other Figma
+layout/typography work, the `useOrganizations` hooks, and publisher names are retained —
+only the font **family** is reverted.
```
cameronapak: issue: remove any and all instances referencing Aktiv Grotesk font because we are not using it

docs/adr/0003-adopt-untitled-serif-via-fonts-api.md
```diff
@@ -0,0 +1,132 @@
...
+## Status
+
+Accepted. Supersedes [ADR-0001](0001-revert-brand-fonts-pending-licensing.md) **in
+part** — for Untitled Serif only. ADR-0001's Aktiv Grotesk revert stands unchanged.
+
+## Context
+
+ADR-0001 reverted both YouVersion brand fonts and set two conditions for bringing one
+back: cleared licensing, and loading through the `/v1/fonts/{font_id}/stylesheet`
+endpoint rather than a hardcoded `@font-face`.
```
cameronapak: [issue](https://conventionalcomments.org/): remove any and all instances of the Aktiv Grotesk font name. The clarity around this is we never actually published with these fonts. I had a friend that was working on tickets, and I imagine, and then he had to rip it out. But I imagine he left some of the ADR documentation in play because he believed that we needed to use Active Grotesque. And anyway, we had to rip it out. And I imagine some of that doc still got pushed. But the truth is we never pushed active grotesque font and we never pushed untitled serif font. And we're pushing untitled serif now and we have permission for that specific font.

packages/core/src/styles/theme.css
```diff
@@ -108,11 +108,15 @@
...
-  /* Brand fonts (Aktiv Grotesk App / Untitled Serif) reverted to Inter / Source
-     Serif 4 pending licensing — see docs/adr/0001-revert-brand-fonts-pending-licensing.md.
-     Brand-font implementation parked on branch feat/youversion-brand-fonts. */
+  /* Untitled Serif is the brand serif, loaded from the gated Fonts API stylesheet
+     endpoint — see docs/adr/0003-adopt-untitled-serif-via-fonts-api.md. Source Serif 4
+     (Google Fonts) stays in the stack as the fallback, so a host that never loads
+     Untitled Serif sees no regression. Aktiv Grotesk App remains reverted pending
+     licensing (docs/adr/0001-revert-brand-fonts-pending-licensing.md), so sans stays Inter.
+     NOTE: the serif stack is duplicated verbatim in packages/ui/src/styles/global.css
+     (@theme inline, --font-serif). Edit both together — font-tokens.test.ts guards drift. */
```
cameronapak, line 116: issue: We have license permission to use Untitled. Please remove references that imply we do not have permissions.

cameronapak:
Ready for review @jaredhightower-youversion or @camrun91 

P.S. Dustin and I are out of office the rest of this week due to being over hours and will resume on Monday

camrun91:
## Review notes — worth fixing before merge

Two items from review that I'd like addressed before this lands:

### [High] Suspending stylesheet not contained

**Files:** `packages/ui/src/components/YouVersionProvider.tsx`, `packages/ui/src/lib/yv-fonts.tsx`

`<YvFonts />` renders a React 19 `<link rel="stylesheet" precedence>` as a sibling of `children` with no local Suspense boundary. When it suspends, the nearest boundary is above the provider in the consumer app — so their entire tree can be held during font fetch.

Combined with no opt-out, every consumer inherits first-nav latency to `api.youversion.com`.

**Suggested fix:** wrap the font link in a local boundary inside the provider:

```tsx
<Suspense fallback={null}>
  <YvFonts … />
</Suspense>
```

That keeps suspension scoped to the provider instead of bubbling up to the app.

---

### [Medium] Changeset is stale vs the branch

**File:** `.changeset/untitled-serif-via-fonts-api.md` (and matching ADR-0004 wording)

The changeset text doesn't match what's actually in the branch:

- It claims the picker reads **"Untitled"**, but the locale string is **`"Untitled Serif"`**.
- It claims `sourceSerifFontName` is gone from private locales, but it's still present in all six locale files.

Those lines ship into published CHANGELOGs, so please update the changeset (and the ADR wording to match) before merge.

camrun91 (changes requested):
See above request of changes.

cameronapak:
@camrun91 really good finds! Updated this PR to reflect those findings/fixes. Please re-review when able.

camrun91 (approved):
Great work LGTM

PR #318, opened 2026-08-04
fix(ui): YPE-3871 - secondary buttons no longer read as disabled

cameronapak:
https://github.com/user-attachments/assets/8e4e821d-601f-428f-a9fd-24bdb6bcef7f

## Summary

Secondary buttons paired `bg-muted` with `text-muted-foreground`, so active controls like the BibleCard version picker read as disabled. The variant now uses the normal foreground color, taking the light-theme label from 5.6:1 to 17.1:1 against the button surface.

YPE-3871 - https://lifechurch.atlassian.net/browse/YPE-3871

## Changes

1. Secondary buttons read as active controls — the variant pairs `bg-muted` with the normal foreground color instead of the muted one.
2. The BibleCard version picker's loading spinner inherits the button label color rather than forcing its own muted one.
3. The BibleCard and Verse of the Day demo pages cap their width, so the cards render in a realistic container instead of stretching to the viewport.

**Start here:** change 1 — it repaints every secondary button, not just BibleCard's.

## Test plan

- `pnpm typecheck` — passed
- `pnpm lint` — passed
- `pnpm --filter @youversion/platform-react-ui test` — 401 passed
- `pnpm --filter @youversion/platform-react-hooks test` — 289 passed
- `@youversion/platform-core` tests not run: `.env.local` is absent locally, and no core files changed
- **Needs manual check:** the version button in light and dark themes, and the RN Expo SDK's BibleCard — the dark theme's `muted-foreground` is near-white, so it hid this bug, and the Expo SDK lives in another repo.

### DIFF
--- a/examples/vite-react/src/pages/BibleCardPage.tsx
+++ b/examples/vite-react/src/pages/BibleCardPage.tsx
@@ -2,7 +2,7 @@ import { BibleCard } from '@youversion/platform-react-ui';
 
 export function BibleCardPage() {
   return (
-    <div className="flex justify-center p-6 md:p-12">
+    <div className="flex justify-center p-6 max-w-screen-sm mx-auto md:p-12">
       <BibleCard reference="JHN.3.16" versionId={3034} showVersionPicker />
     </div>
   );
--- a/examples/vite-react/src/pages/VotdPage.tsx
+++ b/examples/vite-react/src/pages/VotdPage.tsx
@@ -2,7 +2,7 @@ import { VerseOfTheDay } from '@youversion/platform-react-ui';
 
 export function VotdPage() {
   return (
-    <div className="flex flex-col grow items-center gap-8 w-full p-6 md:p-12">
+    <div className="flex flex-col max-w-screen-sm mx-auto grow items-center gap-8 w-full p-6 md:p-12">
       <VerseOfTheDay size="default" versionId={3034} />
       <VerseOfTheDay size="lg" versionId={3034} />
     </div>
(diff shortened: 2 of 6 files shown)


examples/vite-react/src/pages/BibleCardPage.tsx
```diff
@@ -2,7 +2,7 @@ import { BibleCard } from '@youversion/platform-react-ui';
 
 export function BibleCardPage() {
   return (
-    <div className="flex justify-center p-6 md:p-12">
+    <div className="flex justify-center p-6 max-w-screen-sm mx-auto md:p-12">
```
cameronapak, line 5: note: I did this to make it where the Bible card is within container confines, so it's not super long. And the reason why it fits the boundaries of what it's in is because Bible Project needed it to span the full width at the time

examples/vite-react/src/pages/VotdPage.tsx
```diff
@@ -2,7 +2,7 @@ import { VerseOfTheDay } from '@youversion/platform-react-ui';
 
 export function VotdPage() {
   return (
-    <div className="flex flex-col grow items-center gap-8 w-full p-6 md:p-12">
+    <div className="flex flex-col max-w-screen-sm mx-auto grow items-center gap-8 w-full p-6 md:p-12">
```
cameronapak, line 5: Same idea as https://github.com/youversion/platform-sdk-react/pull/318/changes#diff-5c288b499be5f85c9364db8663c5472f20be452ace68dcc174344f589a3b68afR5

packages/ui/src/components/ui/button.test.tsx
cameronapak, line 1: [nitpick](https://conventionalcomments.org/): please remove unnecessary test

bmanquen (approved):

PR #326, opened 2026-08-11
fix: guard storage access by capability instead of typeof window

bmanquen:
- Add `getLocalStorage()`/`getSessionStorage()` to core; route every storage call site in core, hooks, and UI through them
- `typeof window` was the wrong check: React Native sets `global.window === global`, so the guard passed and the next line threw
- Helpers also cover what `typeof localStorage` misses: `null` on Android DOM WebViews, Node's experimental global evaluating to `undefined`, and a `SecurityError` thrown on the property access itself
- Reads resolve to `null` and writes no-op instead of throwing
- `signIn`/`handleAuthCallback` need persistence across the OAuth redirect, so they report the missing capability instead of degrading
- Drop the UI package's module-private `getSessionStorage`, which collided by name with core's

YPE-4815

<h3>Greptile Summary</h3>

The PR replaces direct Web Storage access with capability-checked core helpers and adds throw-safe mutation handling across core, hooks, and UI.
- Reports unavailable storage and rejected critical OAuth writes before redirecting.
- Verifies auth-session writes before completing sign-in or token refresh.
- Makes optional preference, cache, and cleanup operations degrade safely when storage is unavailable.
- Adds coverage for React Native, restricted storage, rejected mutations, and auth persistence failures.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/core/src/web-storage.ts | Introduces capability-checked Web Storage resolution and throw-safe mutation helpers. |
| packages/core/src/YouVersionPlatformConfiguration.ts | Routes ambient authentication and configuration state through the new storage helpers and reports critical persistence outcomes. |
| packages/core/src/Users.ts | Makes OAuth handoff and session persistence failures explicit while keeping post-success cleanup non-throwing. |
| packages/core/src/StorageStrategy.ts | Updates session-storage strategy operations to degrade safely when storage is unavailable or rejects mutations. |
| packages/hooks/src/useYVAuth.ts | Removes the window-based guard and relies on core's safe token accessor. |
| packages/ui/src/components/bible-reader.tsx | Routes reader preference persistence through capability-checked, throw-safe storage helpers. |
| packages/ui/src/components/bible-version-picker.tsx | Makes recent-version storage safe in environments without usable Web Storage. |
| packages/ui/src/lib/pending-highlight.ts | Replaces the UI-local session-storage implementation with core's shared helpers. |

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
  participant App
  participant Core as Core auth flow
  participant Helper as Web Storage helper
  participant Storage as Browser storage
  App->>Core: signIn / handleAuthCallback
  Core->>Helper: getLocalStorage()
  Helper->>Storage: resolve capability
  alt storage unavailable
    Helper-->>Core: null
    Core-->>App: capability error
  else storage available
    Helper-->>Core: Storage
    Core->>Helper: setStorageItem(...)
    Helper->>Storage: setItem(...)
    alt mutation throws
      Helper-->>Core: false
      Core-->>App: persistence error
    else mutation succeeds
      Helper-->>Core: true
      Core-->>App: sign-in result
    end
  end
```

<sub>Reviews (6): Last reviewed commit: ["Merge branch &#39;main&#39; into bm/YPE-4815-cor..."](https://github.com/youversion/platform-sdk-react/commit/44b854d0e9732c16f1a39a0d3965e490ebb2b1a2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52314925)</sub>

**Context used:**


packages/core/src/web-storage.ts
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * Returns a usable `localStorage` implementation, or `null` when none exists.
+ *
+ * `window` and `globalThis` are the same object in browsers and jsdom, but they
+ * can diverge where a DOM is grafted onto a separate `window` (SSR shims, some
+ * test harnesses) — there `window.localStorage` is the real store while the
+ * global is Node's undefined-valued experimental accessor. Check `window` first
+ * so those environments resolve to the store that actually works.
+ */
+export function getLocalStorage(): Storage | null {
```
cameronapak: praise: One capability check plus throw-safe mutators is the right shape.

`getLocalStorage()` / `getSessionStorage()` find a usable store. `setStorageItem` / `removeStorageItem` / `clearStorage` keep later writes from crashing callers.

— Cursor, sent on behalf of Cam.

packages/core/src/Users.ts
```diff
@@ -65,22 +93,36 @@ export function __resetTokenRefreshDedupeForTests(): void {
   inFlightRefresh = null;
 }
 
+/**
+ * Discards everything the OAuth handoff stashed for the redirect round-trip.
+ * Cleanup only — it never throws, so it is safe on both the success and the
+ * failure path of the code exchange.
+ */
+const clearOAuthHandoff = (storage: Storage | null): void => {
```
cameronapak, line 109: praise: `clearOAuthHandoff` removes the duplicate cleanup on success and failure.

Cleanup now goes through `removeStorageItem`, so a throw during cleanup cannot clear a new session.

— Cursor, sent on behalf of Cam.

packages/core/src/Users.ts
```diff
@@ -105,7 +147,7 @@ const stashRequestedPermissions = (state: string, permissions: string[]): void =
     state,
     permissions,
   };
-  localStorage.setItem(REQUESTED_PERMISSIONS_KEY, JSON.stringify(payload));
+  setStorageItem(getLocalStorage(), REQUESTED_PERMISSIONS_KEY, JSON.stringify(payload));
```
cameronapak: issue (blocking): OAuth still mixes hard-fail storage with soft-fail permission stashes.

`signIn` resolves storage with `requireLocalStorage()` and writes the verifier with `persistOrThrow`. Then this helper calls `setStorageItem(getLocalStorage(), ...)` and ignores a failed write.

The same pattern appears for `stashPendingGrantedPermissions` on the callback path. A failed stash drops grants with no error.

Pass the resolved `Storage` into the stash helpers. Or write those keys with `persistOrThrow` when the handoff must keep them.

— Cursor, sent on behalf of Cam.

packages/core/src/Users.ts
```diff
@@ -278,6 +326,7 @@ export class YouVersionAPIUsers {
     state: string | null,
     urlParams: URLSearchParams,
   ): Promise<SignInWithYouVersionResult> {
+    const storage = getLocalStorage();
```
cameronapak: suggestion (non-blocking): Prefer the required store here too.

`handleAuthCallback` already called `requireLocalStorage()`. `exchangeCodeForTokens` probes again with `getLocalStorage()`. Thread the resolved `Storage` from the caller so this path keeps one policy.

— Cursor, sent on behalf of Cam.

packages/core/src/__tests__/Users.test.ts
```diff
@@ -151,6 +151,33 @@ describe('YouVersionAPIUsers', () => {
       );
     });
 
+    it('should report the missing capability when localStorage is unusable', async () => {
```
cameronapak, line 154: todo (test): Add the same missing-storage case for `handleAuthCallback`.

This test covers `signIn`. `handleAuthCallback` also calls `requireLocalStorage()`. A twin test locks that path.

— Cursor, sent on behalf of Cam.

packages/core/src/web-storage.ts
```diff
@@ -0,0 +1,109 @@
...
+    resolveStorage(() => (typeof window !== 'undefined' ? window.localStorage : null)) ??
+    resolveStorage(() => globalThis.localStorage)
+  );
+}
+
+/**
+ * Returns a usable `sessionStorage` implementation, or `null` when none exists.
+ * See {@link getLocalStorage} for why `window` is checked before `globalThis`.
+ */
+export function getSessionStorage(): Storage | null {
```
cameronapak, line 66: nitpick (non-blocking): `getLocalStorage` and `getSessionStorage` share one shape.

A private `getWebStorage(name)` with two thin exports is enough. This is optional.

— Cursor, sent on behalf of Cam.

packages/ui/src/components/bible-version-picker.tsx
```diff
@@ -1,7 +1,12 @@
 import { useTranslation } from 'react-i18next';
 import i18n from '@/i18n';
 import { useControllableState } from '@radix-ui/react-use-controllable-state';
-import type { BibleVersion, Language } from '@youversion/platform-core';
+import {
+  getLocalStorage,
```
cameronapak, line 5: question (non-blocking): Does UI still ban direct `@youversion/platform-core` imports?

`packages/ui/AGENTS.md` says do not import from core except re-exports. These files already import core types. This PR adds storage helpers the same way.

Update the rule, or re-export the helpers from the UI package entry.

— Cursor, sent on behalf of Cam.

packages/ui/src/lib/pending-highlight.ts
```diff
@@ -15,6 +15,8 @@
  * it is a short-lived intent record, discarded on decline, cancel, failure, or
  * successful apply.
  */
+import { getSessionStorage, removeStorageItem, setStorageItem } from '@youversion/platform-core';
```
cameronapak, line 18: praise: Dropping the local `getSessionStorage` helper removes the name clash with core.

UI now uses the shared accessors and throw-safe mutators. That matches the prior-art note in YPE-4815.

— Cursor, sent on behalf of Cam.

cameronapak (commented):
**Review summary**

Thanks for `eaf0538`. The throw-safe mutators and `clearOAuthHandoff` fix the cleanup risk.

Main ask left: keep the hard-fail storage policy consistent for OAuth permission stashes in `Users.ts`.

Inline notes use [Conventional Comments](https://conventionalcomments.org/).

— Cursor, sent on behalf of Cam.

packages/core/src/web-storage.ts
```diff
@@ -0,0 +1,109 @@
...
+ * `QuotaExceededError` — so resolving a store is not enough to make a mutation
+ * safe. Mutate through {@link setStorageItem} / {@link removeStorageItem} /
+ * {@link clearStorage}, never by calling the store's methods directly.
+ */
+
+function resolveStorage(read: () => Storage | undefined | null): Storage | null {
+  try {
+    const storage = read();
+    if (storage && typeof storage.getItem === 'function') {
+      return storage;
```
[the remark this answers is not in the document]
bmanquen: Not fixing. Every read failure we can actually name — Chrome with cookies blocked, Firefox with storage disabled, Node started without --localstorage-file — throws at property access, which resolveStorage's existing try already catches and turns into null. A store that resolves successfully and then throws inside getItem isn't a behavior we can point to in any browser, unlike the write case (Safari private mode reads fine and throws on setItem), which is exactly why the mutation helpers exist and the read path doesn't have an equivalent. Guarding it would cost a probe read on every resolve — or ~14 call-site changes for a getStorageItem wrapper — against a hypothetical, and the probe wouldn't be airtight anyway since a store could still go bad between resolve and read.

packages/core/src/YouVersionPlatformConfiguration.ts
```diff
@@ -32,31 +37,35 @@ export class YouVersionPlatformConfiguration {
...
     accessToken: string | null,
     refreshToken: string | null,
     expiryDate: Date | null,
   ): void {
+    const storage = getLocalStorage();
+    if (!storage) return;
+
     if (accessToken !== null) {
-      localStorage.setItem('accessToken', accessToken);
+      setStorageItem(storage, 'accessToken', accessToken);
```
cameronapak: issue (blocking): OAuth callback can succeed when token writes did not land.

`saveAuthData` uses `setStorageItem` but ignores a `false` return. `exchangeCodeForTokens` / `handleAuthCallback` then return a successful `SignInWithYouVersionResult` while nothing is persisted for reload.

Before this PR, a throwing `setItem` failed the exchange (catch → `clearAuthTokens` + rethrow). That conflicts with the PR's own rule that redirect persistence fails loudly (`persistOrThrow` for verifier / requested permissions).

Hard-fail the OAuth path when access/refresh/expiry (and ideally profile) writes do not land — check the boolean, return it from `saveAuthData`, or persist these keys with `persistOrThrow`. Soft-fail can stay for non-OAuth configuration callers.

Same Greptile P1.

packages/core/src/Users.ts
```diff
@@ -270,13 +351,20 @@ export class YouVersionAPIUsers {
...
    * must dedupe by `code` (see {@link inFlightCodeExchanges}); on failure the
    * partial session is cleared and the error rethrown.
+   *
+   * `storage` is the store the caller already resolved via
+   * {@link requireLocalStorage}, threaded in rather than re-resolved: the
+   * handoff values read here were written to that store, and one policy for
+   * the whole callback path keeps a mid-flow change in the global from
+   * silently reading somewhere else (or nowhere).
    */
   private static async exchangeCodeForTokens(
```
cameronapak, line 369: issue (blocking): This exchange path still trusts `saveAuthData` / `saveUserInfo` as fire-and-forget.

After the code exchange, a rejected localStorage write should fail the callback the same way a missing store does — otherwise Greptile's silent-success case stands.

See the note on `YouVersionPlatformConfiguration.saveAuthData`.

packages/core/src/web-storage.ts
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Writes a value, tolerating a store that throws on mutation (zero quota in
+ * Safari private mode, a full quota, storage disabled mid-session).
+ *
+ * @param storage A resolved store, or `null` — pass `getLocalStorage()` directly.
+ * @returns `true` when the value was actually persisted. Callers that merely
+ *   cache something can ignore it; callers that need the value to survive a
+ *   reload or redirect must not assume the write landed.
+ */
+export function setStorageItem(storage: Storage | null, key: string, value: string): boolean {
```
cameronapak, line 79: praise: Capability resolve + throw-safe mutators is the right split.

`getWebStorage` covers the React Native / null / SecurityError cases. Mutations go through helpers because Safari private mode still throws on write after a successful resolve. Clear docs and tests.

cameronapak (changes requested):
## Standards

**praise:** Follow-up after the earlier review is thorough — resolved `Storage` is threaded through the OAuth handoff, `persistOrThrow` covers the verifier and requested-permissions stash, cleanup goes through `clearOAuthHandoff`, and `packages/ui/AGENTS.md` now states the rule the code already followed.

**suggestion (non-blocking):** Token refresh soft-persists via `saveAuthData` the same way. If you harden the callback path, apply the same check there so a rotated access token cannot look refreshed while storage rejected the write.

No Fowler-smell or repo-standards blockers beyond the Spec item below. Package boundaries, changeset, core/UI AGENTS updates, and capability-checked accessors match `AGENTS.md` / `packages/core/AGENTS.md`.

## Spec

Spec source: PR body + changeset for YPE-4815. Claimed fix: stop `typeof window` storage guards that pass in React Native then throw; route storage through capability helpers; OAuth paths that need redirect persistence fail loudly.

**issue (blocking):** `exchangeCodeForTokens` still calls `YouVersionPlatformConfiguration.saveAuthData(...)` (and `saveUserInfo`) without checking whether those writes landed. `setStorageItem` returns `false` on quota/disabled-store failures, but `saveAuthData` discards that and the callback returns a successful `SignInWithYouVersionResult`. Pre-PR, a throwing `setItem` entered the catch and cleared the session. That matches Greptile's open P1 on silent OAuth token writes.

Agree with Greptile on this one. Agree with Brenden (and Greptile's withdrawal) that a `getStorageItem` wrapper for hypothetical post-resolve `getItem` throws is not worth the cost.

Prior review asks look addressed: stash helpers take resolved storage / `persistOrThrow` where needed, `exchangeCodeForTokens` no longer re-probes, `handleAuthCallback` missing-storage test exists, getters collapsed, UI AGENTS updated.

packages/core/src/Users.ts
```diff
@@ -343,26 +431,31 @@ export class YouVersionAPIUsers {
...
+      const profilePersisted = YouVersionPlatformConfiguration.saveUserInfo({
         id: result.yvpUserId,
         name: result.name,
         email: result.email,
         avatar_url: result.profilePicture,
       });
 
+      // Every later read of the session goes back to storage, so a rejected
+      // write means there is no session to return — fail into the catch below.
+      if (!tokensPersisted || !profilePersisted) {
```
cameronapak, line 451: praise: This closes the prior blocking gap.

Checking `tokensPersisted` / `profilePersisted` means the callback can no longer report success when session writes failed. Matches the PR's hard-fail policy for redirect persistence.

packages/core/src/Users.ts
```diff
@@ -4,6 +4,34 @@ import { YouVersionPlatformConfiguration } from './YouVersionPlatformConfigurati
...
+
+/**
+ * Writes a value the OAuth callback cannot proceed without. A resolvable store
+ * can still reject writes (Safari private mode), and silently losing the
+ * verifier here would surface after the redirect as the far more confusing
+ * "Missing required authentication parameters" — so fail now, before redirecting.
+ */
+const persistOrThrow = (storage: Storage, key: string, value: string): void => {
+  if (!setStorageItem(storage, key, value)) {
+    throw new Error(MISSING_LOCAL_STORAGE_MESSAGE);
```
cameronapak: nitpick (non-blocking): Error string says storage is "not available" even when the store resolved but `setItem` failed (Safari private / quota).

A slightly different message for write rejection would make the failure mode obvious in the wild.

packages/core/src/YouVersionPlatformConfiguration.ts
```diff
@@ -32,44 +37,66 @@ export class YouVersionPlatformConfiguration {
...
+   * Persists the session tokens; a `null` value clears the corresponding key.
+   *
+   * @returns `true` when every value it was asked to persist actually landed.
+   *   The session lives entirely in storage — the {@link accessToken},
+   *   {@link refreshToken}, and {@link tokenExpiryDate} getters read straight
+   *   back out of it — so a caller reporting a successful sign-in must not do
+   *   so on `false`. Clearing always reports `true`: removal is best-effort,
+   *   and every reader tolerates a value that outlived its removal.
+   */
   public static saveAuthData(
```
cameronapak, line 55: praise: Returning `boolean` from `saveAuthData` is the right API for callers that must survive a reload.

Soft-fail stays available to ignore; OAuth/refresh can hard-fail. Clearer than swallowing `setStorageItem`'s `false`.

cameronapak (dismissed):
## Standards

**praise:** Prior soft-fail on `saveAuthData` is cleaned up in the helpers and call sites — boolean returns, and OAuth/refresh paths check them. Touched call sites go through capability helpers / throw-safe mutators; UI imports from the package entry; changeset present.

**nitpick (non-blocking, Mysterious Name):** `persistOrThrow` reuses `MISSING_LOCAL_STORAGE_MESSAGE` (“…not available…”) when the store resolved but `setItem` failed (Safari quota). Worth a clearer message so Safari-private failures don’t look like “no storage”.

**thought (non-blocking):** Root `AGENTS.md` still says core must not use DOM APIs; the `web-storage.ts` carve-out lives only in `packages/core/AGENTS.md`. Package doc is the intentional override — consider syncing the root line later.

**thought (non-blocking, Duplicated Code):** `bible-reader.tsx` still has three near-identical hydrate/persist effect pairs — pre-existing shape, just routed through the helpers.

Hard standards violations: none. Smell baseline items above are judgement calls.

## Spec

Spec sources: YPE-4815 + PR body/changeset + prior REQUEST_CHANGES + Greptile.

**praise:** Prior blocking ask is fixed at this head — `saveAuthData`/`saveUserInfo` return `boolean`; `exchangeCodeForTokens` / refresh throw when persistence fails. YPE-4815 sites and capability guards look complete. Greptile P1 silent token writes and throwable mutations addressed; agree with Brenden/Greptile withdrawal on hypothetical post-resolve `getItem` throws.

**note:** `packages/ui/AGENTS.md` rewrite is outside the ticket’s four named sites but matches the PR’s own docs commit and the imports this change needs — not silent creep.

Missing/wrong requirements: none material remaining.

**Summary:** Standards 0 hard / a few non-blocking judgement calls (worst: unclear `persistOrThrow` error string). Spec 0 findings (prior blocker resolved).

cameronapak (approved):

PR #335, opened 2026-08-18
feat(ui): paint host highlights on BibleTextView, BibleCard, and VerseOfTheDay (YPE-5064)

Dustin-Kelley:
## Summary
- Add an optional paint-only `highlights` prop on `BibleTextView`, `BibleCard`, and `VerseOfTheDay` so hosts that already own highlight data (React Native / Expo DOM) can render it without a fetch or verse-action UI (YPE-4492).
- Project host `Highlight[]` onto the displayed chapter via `deriveHighlightedVerses`; Verse of the Day also filters to today's passage so a host-supplied superset cannot paint verses that are not shown.
- Latch paint mode at first mount (`highlights` present vs omitted) and prefer `highlights` over `highlightedVerses` when both are passed, with a changeset for the UI package.

## Test plan
- [x] Confirm omitting `highlights` leaves existing `highlightedVerses` / reader behavior unchanged.
- [x] Pass `highlights` into `BibleTextView`, `BibleCard`, and `VerseOfTheDay` and confirm matching verses paint with the expected colors.
- [x] Pass `[]` and confirm nothing paints (controlled empty, not a fallback to `highlightedVerses`).
- [x] On Verse of the Day, pass highlights for other verses/chapters and confirm only today's passage paints.
- [x] Storybook: check the new highlight stories for `BibleTextView`, `BibleCard`, and `VerseOfTheDay`.

Made with [Cursor](https://cursor.com)

<h3>Greptile Summary</h3>

The PR adds controlled and self-contained highlight painting to `BibleTextView`, `BibleCard`, and `VerseOfTheDay`.
- Projects host or fetched highlight rows onto the currently rendered scripture passage.
- Clips Verse of the Day and verse-range surfaces to the displayed passage.
- Shares first-mount controlled-mode latching across scripture components.
- Adds component, hook, projection, and Storybook coverage plus a UI-package changeset.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/ui/src/components/verse.tsx | Adds the public highlight prop, controlled-mode latch, rendered-passage scoping, and the shared paint hook to BibleTextView. |
| packages/ui/src/components/verse-of-the-day.tsx | Adds host and self-contained highlight painting while clipping rows to the resolved daily passage. |
| packages/ui/src/components/bible-card.tsx | Exposes and forwards the optional highlight prop to BibleTextView. |
| packages/ui/src/lib/highlight-projection.ts | Adds passage clipping and retained-passage scope helpers for projecting highlight rows safely. |
| packages/ui/src/lib/use-scripture-highlight-paint.ts | Centralizes controlled, reader-owned, and permission-gated self-contained highlight painting. |
| packages/ui/src/lib/use-highlights-controlled-latch.ts | Extracts the shared first-mount controlled-mode latch and development warning behavior. |

<h3>Flowchart</h3>

```mermaid
flowchart TD
  Surface[BibleTextView / BibleCard / VerseOfTheDay]
  Latch{highlights present at first mount?}
  Host[Use host Highlight array]
  Reader{highlightedVerses supplied?}
  Existing[Use reader verse-color map]
  Eligible{Signed in, permission granted, feature live?}
  Fetch[Fetch chapter highlights]
  Empty[Paint nothing]
  Clip[Clip to rendered passage]
  Project[Project rows to verse-color map]
  Render[Render highlighted verses]

  Surface --> Latch
  Latch -->|Yes| Host
  Latch -->|No| Reader
  Reader -->|Yes| Existing
  Reader -->|No| Eligible
  Eligible -->|Yes| Fetch
  Eligible -->|No| Empty
  Host --> Clip
  Fetch --> Clip
  Clip --> Project
  Project --> Render
  Existing --> Render
  Empty --> Render
```

<sub>Reviews (16): Last reviewed commit: ["test(ui): drop the extra host-highlights..."](https://github.com/youversion/platform-sdk-react/commit/bef578c9bca68c4e8b0bef27465a711e4e24c8b7) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54355652)</sub>

**Context used:**


cameronapak (commented):
I'm Cursor, sent on behalf of Cam.

Two-axis review of this PR vs `main` against YPE-5064. Inline notes use [Conventional Comments](https://conventionalcomments.org/).

**Standards:** 3 judgement calls. No hard documented-standard breach. Worst: duplicated latch/warn vs `BibleReader.Root`.

**Spec:** 1 finding. Worst: VerseOfTheDay latches on child mount after load, not on VOTD first mount.

packages/ui/src/components/verse-of-the-day.tsx
```diff
@@ -250,6 +276,7 @@ export function VerseOfTheDay({
                 loading: isLoading,
                 error: errorPassage || errorVerseOfTheDay || null,
               }}
+              highlights={paintHighlights}
```
cameronapak: issue (blocking): VerseOfTheDay does not latch `highlights` at its own first mount.

YPE-5064: "Latching matches `BibleReader.Root` (YPE-3705)." Root latches `highlights !== undefined` on first render. This component only mounts `BibleTextView` after `!isLoading`, so the latch happens on the child's first mount.

If the first VOTD render omits `highlights` and the host passes an array before load finishes, `BibleTextView` first sees a present prop and paints. Root would ignore that flip.

The JSDoc ("A first render of `undefined` will never paint later data") is false in that window. Latch on `VerseOfTheDay` itself and forward the latched value.

packages/ui/src/components/verse.tsx
```diff
@@ -556,6 +559,15 @@ export type BibleTextViewProps = {
   selectedVerses?: number[];
   onVerseSelect?: (verses: number[]) => void;
   highlightedVerses?: Record<number, string>;
+  /**
+   * Host-supplied highlights in the core API shape. Presence of this prop
+   * (including `[]`) latches paint-only mode at first mount: verses are
```
cameronapak: suggestion (non-blocking): Use Controlled mode, or update CONTEXT.md.

CONTEXT.md says use terms exactly. Controlled mode is host `highlights: Highlight[]`, latched at first mount, pure projection. This JSDoc (and the changeset) coins "paint-only" / "controlled paint" instead. If this is the same posture, use Controlled mode. If it is a distinct posture, add it to the glossary.

packages/ui/src/components/verse.tsx
```diff
@@ -586,6 +599,36 @@ export const BibleTextView = forwardRef<HTMLDivElement, BibleTextViewProps>(
     const providerTheme = useTheme();
     const currentTheme = theme || providerTheme;
 
+    // Latched at first mount: a transient `undefined` on a controlled view must
+    // render as "no highlights", never fall through to `highlightedVerses`.
+    const isHighlightsControlledRef = useRef(highlights !== undefined);
```
cameronapak: suggestion (non-blocking): Extract the latch/warn helper.

This repeats `BibleReader.Root`. The ADR endorses the behavior, not the copy. One helper keeps the warn copy and latch rules from drifting.

packages/ui/src/components/verse-of-the-day.tsx
```diff
@@ -65,8 +66,30 @@ export type VerseOfTheDayProps = {
...
+   * Use for React Native or Expo DOM hosts that already own highlight data
+   * and just need it rendered. Mode is latched at first mount: omit the prop
+   * for no highlights; pass `[]` for none yet. A first render of `undefined`
+   * will never paint later data. Nothing paints until today's passage is
+   * known; overlapping ranges are clipped to that passage.
+   */
+  highlights?: Highlight[];
 };
 
+function paintHighlightsForPassage(
```
cameronapak: nitpick (non-blocking): This helper filters; it does not paint.

Inline it, or name it for the clip (`clipHighlightsToPassage`).

packages/ui/src/lib/highlight-projection.ts
```diff
@@ -44,6 +45,94 @@ export function expandPassageId(passageId: string): ExpandedPassageId | null {
...
+ * against the requested USFM in that window paints destination-chapter
+ * highlights onto the old passage's same-numbered verse wrappers.
+ *
+ * When the retained passage is a different chapter, this returns that
+ * chapter so the old HTML keeps its own highlights. When it is the same
+ * chapter but still loading, this returns `null`: the retained HTML may
+ * belong to the previous version, and `BiblePassage` has no version id to
+ * confirm a match.
+ */
+export function chapterScopeForHighlightPaint(options: {
```
cameronapak, line 85: praise: Projecting against the rendered passage, not the requested one.

That matches `useApiData` keeping the previous passage on screen. Avoids painting destination colors onto the old wrappers.

cameronapak (commented):
## Two-axis review: Standards and Spec

Fixed point: `e10aeb4`, the merge-base with `main`. Spec: [YPE-5064](https://lifechurch.atlassian.net/browse/YPE-5064), a sub-task of [YPE-4492](https://lifechurch.atlassian.net/browse/YPE-4492).

The projection work is correct. I checked each rule the ticket named, and the tests cover them.

**Spec: 4 findings.** One blocks. Commit `3a39d9e` adds a self-contained fetch. The ticket says "No fetch inside these components" and "Prop omitted -> no host highlights (and no fetch in this ticket)". The parent story defers that work until the reader flow settles the cache rules.

**Standards: 7 findings.** The worst is shared mutable mock state across test cases. `docs/testing.md` lists "No shared mutable state across cases" as a must.

Findings are inline. Each one names its source: the ticket, or the repo document. Comments follow [conventional comments](https://conventionalcomments.org/).

packages/ui/src/lib/use-scripture-highlight-paint.ts
```diff
@@ -0,0 +1,94 @@
...
+  const canFetch =
+    !isHighlightsControlled &&
+    highlightedVerses === undefined &&
+    isAuthenticated &&
+    YouVersionPlatformConfiguration.hasPermission('highlights') &&
+    isHighlightsLive() &&
+    versionId > 0 &&
+    chapterUsfm.length > 0;
+
+  const { highlights: fetchedHighlights } = useHighlights(
```
cameronapak, line 96: **issue (blocking):** this fetch is outside the scope of YPE-5064

The ticket says: "No fetch inside these components, no verse-tap, no create." It also says: "Prop omitted -> no host highlights (and no fetch in this ticket)."

Commit `3a39d9e` adds the opposite. When the prop is omitted and the user is signed in with the `highlights` permission, this hook sends a GET.

The parent story YPE-4492 defers the work: "Needs the reader flow (YPE-3710 / YPE-3711) settled first, so the paint pipeline and cache scope rules are established and not re-invented here."

That dependency is load-bearing here. `useApiData` has no module cache and no in-flight dedupe. Each mounted `BibleCard` and `VerseOfTheDay` therefore sends its own GET per chapter. A page with six cards sends six identical requests.

The changeset also advertises auto-paint as the public default, and `CONTEXT.md` redefines Self-contained mode to cover all three components. That is a product decision, not this sub-task.

Two ways forward:

1. Remove the self-contained fetch. Ship controlled paint only, as the ticket defines it. File the fetch as its own ticket.
2. Keep the fetch, and first agree the scope change on YPE-5064. Then settle the cache rules.

Option 1 is cleaner. It keeps this PR reviewable against a spec that matches it.

packages/ui/src/components/verse.tsx
```diff
@@ -556,6 +561,24 @@ export type BibleTextViewProps = {
...
+   * (including `[]`) latches **controlled mode** at first mount: verses are
+   * projected from this list and nothing is fetched. Use this for React Native
+   * or Expo DOM hosts that own highlight data (keep the token out of the
+   * WebView) — pass `[]` while loading or signed out so the WebView does not
+   * fetch.
+   *
+   * Omit the prop for **self-contained** paint: when the
+   * user is signed in, has granted the `highlights` permission, and highlights
+   * are live, matching verses are fetched and painted. A first render of
+   * `undefined` latches self-contained (fetch when eligible), not "never paint".
```
cameronapak, line 577: **issue:** the omitted-prop latch no longer matches `BibleReader.Root`

YPE-5064 says: "Latching matches `BibleReader.Root` (YPE-3705): Prop omitted -> no host highlights."

This line defines an omitted prop as "fetch when eligible". That is a different posture from the reader. The difference follows from the fetch, and it goes away when the fetch goes away.

packages/ui/src/components/verse-of-the-day.tsx
```diff
@@ -149,6 +180,14 @@ export function VerseOfTheDay({
 
   const isLoading = loadingPassage || loadingVerseOfTheDay || loadingVersion;
 
+  // Latched at this component's first mount, not at BibleTextView's. VOTD
+  // delays mounting the child until load finishes; latching only there would
+  // treat a highlights array that arrived during load as "present on first
+  // mount" and paint, which BibleReader.Root would ignore.
+  const isHighlightsControlled = useHighlightsControlledLatch(highlights, 'VerseOfTheDay');
+  const hostHighlights = isHighlightsControlled ? (highlights ?? []) : undefined;
+  const clippedHighlights = clipHighlightsToPassage(hostHighlights, data?.passage_id);
```
cameronapak, line 189: **issue:** the clip runs in controlled mode only

`hostHighlights` is `undefined` in self-contained mode. `clipHighlightsToPassage` returns `undefined` for that input, so fetched rows reach `deriveHighlightedVerses` unclipped.

The clip is not cosmetic. The test at `verse-of-the-day.test.tsx:304` renders verse 16 and verse 17, and today's passage is verse 16 only. A fetched chapter highlight on verse 17 then paints verse 17. That verse is on screen, but it is not today's verse. The controlled path blocks this. The fetched path does not.

If the fetch stays, clip both inputs.

packages/ui/src/components/bible-card.tsx
```diff
@@ -24,6 +24,19 @@ export type BibleCardProps = {
...
+   * already own highlight data — pass `[]` while loading or signed out so the
+   * WebView does not fetch.
+   *
+   * Omit the prop for self-contained paint: when the user is signed
+   * in, has granted the `highlights` permission, and highlights are live, the
+   * card fetches and paints matching verses. A first render of `undefined`
+   * latches self-contained (fetch when eligible). Mode is latched at first
+   * mount.
+   */
+  highlights?: Highlight[];
```
cameronapak, line 43: **question:** does `BibleCard` need the same clip as `VerseOfTheDay`?

`VerseOfTheDay` clips host rows to the displayed passage. This component forwards `highlights` unchanged, and its own stories render a range (`LUK.1.39-45`).

The `VerseOfTheDay` test HTML carries the verse after today's verse. If range HTML behaves the same way here, a host row like `LUK.1.39-50` paints verses the card does not claim to show.

If the card cannot hit that case, one sentence in this doc comment that says why makes the difference readable.

packages/ui/src/components/bible-card.test.tsx
```diff
@@ -252,3 +270,116 @@ describe('BibleCard - onFootnotePress callback', () => {
...
+        <Providers>
+          <BibleCard reference="JHN.1" versionId={111} highlights={[]} />
+        </Providers>,
+      );
+      expect(vi.mocked(useHighlights)).toHaveBeenCalledWith(
+        { version_id: 111, passage_id: 'JHN.1' },
+        { enabled: false },
+      );
+      expect(getVerseEl(hostEmpty.container, 2).style.backgroundColor).toBe('');
+    } finally {
```
cameronapak: **issue:** this `finally` block restores state that the cases share

`docs/testing.md` lists two musts: "No shared mutable state across cases" and "No `beforeEach`/`afterEach`".

`vi.mocked(useHighlights).mockReturnValue(...)` writes module mock state. This block resets it so that later cases still pass. The reset is the leak made visible. `verse.test.tsx` and `verse-of-the-day.test.tsx` repeat the pattern.

Set the return value inside each case, from a factory that returns a ready-to-run object. Then no case depends on cleanup from another case.

packages/ui/src/test/highlights-test-utils.tsx
```diff
@@ -8,6 +8,33 @@ export function collection(data: Highlight[]): Collection<Highlight> {
   return { data, next_page_token: null };
 }
 
+/** Multi-verse YVDOM used by controlled-mode highlight tests (`.yv-v[v]` wrappers). */
```
cameronapak, line 42: **suggestion:** add the `useHighlights` stub factory here

The same six-field object appears about nine times across four test files:

```ts
{ highlights: { data: [], next_page_token: null }, loading, error, refetch, createHighlight, deleteHighlight }
```

`use-scripture-highlight-paint.test.tsx` already has the shape in `mockUseHighlights`. This PR edits this shared file already.

`docs/testing.md` asks for "inline setup or call factories that return ready-to-run objects". A factory here serves both that rule and the one above.

packages/ui/src/components/verse.test.tsx
```diff
@@ -1242,3 +1254,265 @@ describe('BibleTextView - onFootnotePress callback', () => {
...
+      .mockReturnValue(true);
+
+    try {
+      const { container } = render(
+        <Providers>
+          <BibleTextView reference="JHN.1" versionId={111} passageState={passageState} />
+        </Providers>,
+      );
+
+      expect(vi.mocked(useHighlights)).toHaveBeenCalledWith(
```
cameronapak: **suggestion:** assert the painted output, not the hook call

`docs/testing.md` says to assert behavior and stable contracts. This assertion pins the argument shape of an internal call. A host never sees that shape. The same case already asserts the paint, which is the contract a host does see.

``expect.stringContaining('`highlights` wins')`` elsewhere in this file has the same problem. It pins warning prose, so a copy edit breaks a passing test.

packages/ui/src/components/verse.tsx
```diff
@@ -586,6 +610,22 @@ export const BibleTextView = forwardRef<HTMLDivElement, BibleTextViewProps>(
     const providerTheme = useTheme();
     const currentTheme = theme || providerTheme;
 
+    // Latched at first mount. Controlled (prop present) never falls through to
+    // `highlightedVerses` or the self-contained fetch.
+    const isHighlightsControlled = useHighlightsControlledLatch(highlights, 'BibleTextView');
+    const didWarnDualHighlightPropsRef = useRef(false);
```
cameronapak, line 618: **nitpick:** the warn-once shape is now written twice

`useHighlightsControlledLatch` already holds a warn-once ref, the `IS_PRODUCTION` guard, and a `console.warn`. This block repeats that shape for the dual-prop warning, one line after the extraction.

A second argument on the latch hook, or a small `useWarnOnce`, keeps the rule and the copy in one place.

packages/ui/src/components/verse.stories.tsx
```diff
@@ -193,6 +193,17 @@ export const FullChapter: Story = {
   },
 };
 
+export const PaintOnlyHighlights: Story = {
```
cameronapak: **nitpick:** this story name uses the term the PR retired

Commit `b3f458b` renames the concept to Controlled mode. `CONTEXT.md` opens with: "Terms here are canonical: code, docs, and conversation should use them exactly."

`PaintOnlyHighlights` and `.changeset/paint-only-highlights.md` keep the old term. `packages/ui/AGENTS.md` calls stories the live reference for props and states, so this name is public.

`ControlledHighlights` matches the glossary.

packages/ui/src/lib/highlight-projection.ts
```diff
@@ -44,6 +45,94 @@ export function expandPassageId(passageId: string): ExpandedPassageId | null {
...
+      renderedScope.chapter === requestedScope.chapter;
+    if (sameChapter) return null;
+  }
+
+  return paintScope;
+}
+
+/**
+ * Keeps host highlights whose expanded verses intersect the displayed passage,
+ * rewriting each kept row's `passage_id` to that intersection.
```
cameronapak, line 145: **nitpick:** the doc says one row, the code can emit several

"rewriting each kept row's `passage_id` to that intersection" reads as one row in, one row out. The loop calls `buildPassageIds` and pushes one row per contiguous run. A host row with a gap in it becomes two rows.

packages/ui/src/lib/use-scripture-highlight-paint.ts
```diff
@@ -0,0 +1,94 @@
...
+  const isAuthenticated = Boolean(authContext?.userInfo);
+  const chapterUsfm = chapterScope ? `${chapterScope.book}.${chapterScope.chapter}` : '';
+  const canFetch =
+    !isHighlightsControlled &&
+    highlightedVerses === undefined &&
+    isAuthenticated &&
+    YouVersionPlatformConfiguration.hasPermission('highlights') &&
+    isHighlightsLive() &&
+    versionId > 0 &&
+    chapterUsfm.length > 0;
```
cameronapak: **nitpick:** `chapterUsfm.length > 0` repeats the `chapterScope` check

`chapterUsfm` is empty exactly when `chapterScope` is `null`. `chapterScope !== null` reads as the same condition, without the round-trip through a string.

packages/ui/src/lib/highlight-projection.ts
```diff
@@ -44,6 +45,94 @@ export function expandPassageId(passageId: string): ExpandedPassageId | null {
...
+    : null;
+  const requestedScope = parseChapterScopeFromUsfm(options.requestedUsfm);
+  const paintScope = renderedScope ?? requestedScope;
+  if (!paintScope) return null;
+
+  if (options.loading && renderedScope && requestedScope) {
+    const sameChapter =
+      renderedScope.book === requestedScope.book &&
+      renderedScope.chapter === requestedScope.chapter;
+    if (sameChapter) return null;
```
cameronapak, line 101: **note:** paint blanks during a same-chapter refetch

This branch returns `null` while a same-chapter fetch is in flight. Paint drops to `{}`, then returns after load. A `BibleCard` version switch therefore flashes the highlights off.

The comment above gives the reason, and the trade is sound: a wrong paint is worse than no paint. I am flagging it so that the behavior is a known one, not a bug report later.

packages/ui/src/lib/highlight-projection.ts
```diff
@@ -44,6 +45,94 @@ export function expandPassageId(passageId: string): ExpandedPassageId | null {
...
+
+/**
+ * Keeps host highlights whose expanded verses intersect the displayed passage,
+ * rewriting each kept row's `passage_id` to that intersection.
+ *
+ * A host range like `JHN.3.16-18` against today's `JHN.3.16` becomes
+ * `JHN.3.16`, so {@link deriveHighlightedVerses} cannot paint verses that are
+ * not shown. Chapter-scope and other unexpandable `passage_id`s are dropped.
+ */
+export function filterHighlightsForPassage(
```
cameronapak, line 152: **praise:** the projection rules hold

I checked each rule the ticket named. Ranges expand through `expandPassageId`. Rows for another version or another chapter are skipped. `normalizeHighlightHex` drops an invalid hex. The tests cover all three.

packages/ui/src/lib/use-scripture-highlight-paint.ts
Dustin-Kelley: Scope has changed lol

Dustin-Kelley: Keeping the self-contained fetch. We changed the scope of this PR: omit `highlights` now means signed in + highlights permission + live → fetch and paint. React Native / Expo still pass `[]` or rows from first mount so the WebView never fetches.

Duplicate GETs per mounted card are accepted here (`useApiData` has no module cache). Sharing cache can follow without changing this contract.

packages/ui/src/components/verse.tsx
Dustin-Kelley: Intentional with that default. Omit latches self-contained (fetch when eligible), not "never paint". `BibleReader` still always passes `highlightedVerses`, so it does not take this path. Native hosts must pass `[]` from first mount.

packages/ui/src/lib/highlight-projection.ts
Dustin-Kelley: Agreed — blank paint during a same-chapter refetch is the trade we want. Wrong colors on retained HTML would be worse.

packages/ui/src/lib/highlight-projection.ts
Dustin-Kelley: Thanks — that projection behavior is staying.

packages/ui/src/components/verse-of-the-day.tsx
Dustin-Kelley: Good catch. Fetched rows now clip to the displayed verse/range USFM the same way host rows do, so a chapter GET cannot paint neighboring verses that happen to be in the HTML.

packages/ui/src/components/bible-card.tsx
Dustin-Kelley: Yes for verse and range `reference`s. `BibleTextView` now clips host and fetched rows to that USFM, so a highlight that runs past the card does not paint extra verses. Chapter-scope references still paint the whole chapter, which is what the card claims to show.

packages/ui/src/components/bible-card.test.tsx
Dustin-Kelley: Moved the stub onto a per-case factory (`stubUseHighlights`) so later tests do not depend on a `finally` restore.

packages/ui/src/test/highlights-test-utils.tsx
Dustin-Kelley: Added `stubUseHighlights` in `highlights-test-utils` and pointed the paint tests at it.

packages/ui/src/components/verse.test.tsx
Dustin-Kelley: Dropped the `useHighlights` argument assertions and the warning-copy match. The case still asserts the painted fills.

packages/ui/src/components/verse.tsx
Dustin-Kelley: Extracted `warnOnce` next to the latch helper. The dual-prop warning uses it too.

packages/ui/src/components/verse.stories.tsx
Dustin-Kelley: Renamed the story to `ControlledHighlights` and the changeset file away from paint-only.

packages/ui/src/lib/highlight-projection.ts
Dustin-Kelley: Updated the doc: a gapped intersection becomes one row per contiguous run.

packages/ui/src/lib/use-scripture-highlight-paint.ts
Dustin-Kelley: `canFetch` now checks `chapterScope !== null` instead of the derived string length.

packages/ui/src/test/highlights-test-utils.tsx
```diff
@@ -8,6 +12,53 @@ export function collection(data: Highlight[]): Collection<Highlight> {
   return { data, next_page_token: null };
 }
 
+/**
+ * Ready-to-run `useHighlights` stub. Each case that needs a specific return
+ * calls this instead of writing module mock state and restoring it later.
+ */
+export function stubUseHighlights(
```
cameronapak: **issue (blocking):** `stubUseHighlights` writes the module mock and leaves it in place.

`docs/testing.md` says there is no shared mutable state across cases. A later `it` that does not call the stub can inherit the last return.

Restore the mock at the end of the case, or make the factory return a local fake that the case passes in.

packages/ui/src/components/bible-card.test.tsx
```diff
@@ -252,3 +277,100 @@ describe('BibleCard - onFootnotePress callback', () => {
     expect(data.notes[0]).toContain('Or understood');
   });
 });
+
+describe('BibleCard - host highlights (controlled mode)', () => {
```
cameronapak: **nitpick (non-blocking):** `docs/testing.md` asks for one optional top-level `describe` per module.

This extra group is readable. It is still a second describe in an edited file.

packages/ui/src/lib/use-scripture-highlight-paint.ts
```diff
@@ -0,0 +1,116 @@
...
+    versionId > 0 &&
+    chapterScope !== null;
+
+  const { highlights: fetchedHighlights } = useHighlights(
+    { version_id: versionId, passage_id: chapterUsfm },
+    { enabled: canFetch },
+  );
+
+  return useMemo(() => {
+    if (isHighlightsControlled) {
```
cameronapak, line 102: **note (non-blocking):** Controlled `highlights` wins when `highlightedVerses` is also passed.

The ticket does not define that precedence. `BibleReader` never passes both, so the reader map is safe.

packages/ui/src/lib/highlight-projection.ts
```diff
@@ -44,6 +45,95 @@ export function expandPassageId(passageId: string): ExpandedPassageId | null {
...
+    : null;
+  const requestedScope = parseChapterScopeFromUsfm(options.requestedUsfm);
+  const paintScope = renderedScope ?? requestedScope;
+  if (!paintScope) return null;
+
+  if (options.loading && renderedScope && requestedScope) {
+    const sameChapter =
+      renderedScope.book === requestedScope.book &&
+      renderedScope.chapter === requestedScope.chapter;
+    if (sameChapter) return null;
```
cameronapak, line 101: **note (non-blocking):** This returns `null` when the same chapter is still loading, so paint goes blank.

YPE-5064 asks to project onto the displayed USFM. Blanking during a same-chapter refetch is extra. Name it in the ticket if you keep it.

cameronapak (changes requested):
## Why

YPE-5064 asks for paint-only highlights on `BibleTextView`, `BibleCard`, and `VerseOfTheDay`. Two modes lock on first mount. Omit the prop to fetch when the user is signed in with permission. Pass the array to paint that data with no fetch.

Parent [YPE-4492](https://lifechurch.atlassian.net/browse/YPE-4492) wants highlights on every scripture surface, not only the reader.

```text
omit highlights
  -> self-contained fetch (useHighlights)
  -> clip to displayed USFM
  -> paint

pass highlights / []
  -> project host array
  -> no fetch
  -> clip to displayed USFM
  -> paint

BibleReader
  -> always pass highlightedVerses (even {})
  -> never fetch here
```

## Standards

**Hard**

- `docs/testing.md`: "No shared mutable state across cases." `stubUseHighlights` writes the module `useHighlights` mock and does not restore it. A later case that does not call the stub can inherit the last return.
- `docs/testing.md`: "Flat tests: one optional top-level describe." New grouping describes were added in BibleCard, VerseOfTheDay, BibleTextView, and projection tests.

**Judgement**

- Duplicated Code: omit-vs-empty is re-tested in the hook and in each component.
- Middle Man: VerseOfTheDay `clipHighlightsToPassage` wraps `filterHighlightsForPassage`. The paint hook clips again.
- Mysterious Name: `useScriptureHighlightPaint` vs the glossary terms.

## Spec

**(a) Missing / partial:** none.

**(b) Scope creep**

- Dual-prop rule: `highlights` wins over `highlightedVerses`. The ticket does not define precedence. The reader never passes both, so this does not steal the reader map.
- `chapterScopeForHighlightPaint` returns `null` (blank paint) when the same chapter is still loading. The ticket asks to project onto the displayed USFM.

**(c) Implemented but wrong:** none.

---

Standards: 2 hard, 3 judgement (worst: shared `useHighlights` mock). Spec: 0 missing, 2 scope-creep (worst: dual-prop precedence).

.changeset/scripture-surface-highlights.md
cameronapak, line 1: praise: love to see it!

packages/ui/src/components/verse-of-the-day.tsx
```diff
@@ -65,8 +67,36 @@ export type VerseOfTheDayProps = {
...
+   * mount, including while today's verse is still loading. Nothing paints
+   * until today's passage is known; overlapping ranges are clipped to that
+   * passage.
+   *
+   * Omit the prop for self-contained paint: when the user is signed
+   * in, has granted the `highlights` permission, and highlights are live,
+   * matching verses are fetched and painted. A first render of `undefined`
+   * latches self-contained (fetch when eligible), not "never paint".
+   */
+  highlights?: Highlight[];
```
cameronapak, line 84: praise: while verbose, I'm a big fan of JSDoc, so I appreciate this!

packages/ui/src/lib/highlight-projection.test.ts
cameronapak, line 1: praise: thanks for creating tests for this new helper!

packages/ui/src/lib/use-highlights-controlled-latch.ts
```diff
@@ -0,0 +1,32 @@
+import { useRef } from 'react';
+import { IS_PRODUCTION } from '@/lib/constants';
+
+/** Logs `message` once per ref in development. */
+export function warnOnce(didWarnRef: { current: boolean }, message: string): void {
+  if (IS_PRODUCTION || didWarnRef.current) return;
+  didWarnRef.current = true;
+  console.warn(message);
+}
```
cameronapak, line 9: thought (non-blocking): This seems great and I can totally see us using this again in other places.

cameronapak (approved):
## Why

Re-review of [YPE-5064](https://lifechurch.atlassian.net/browse/YPE-5064) at `bef578c`. Paint-only highlights on `BibleTextView`, `BibleCard`, and `VerseOfTheDay`. Two modes lock on first mount.

Prior blocking item: `stubUseHighlights` left the module mock in place. That is fixed. The factory returns a restore, and the cases call it in `finally`.

```text
omit highlights -> fetch when signed in + permission -> clip -> paint
pass highlights / [] -> project host array -> no fetch -> clip -> paint
BibleReader -> highlightedVerses even {} -> never fetch here
```

## Standards

**Prior hard, now fixed**

- Shared `useHighlights` mock: restore runs in `finally`.
- Extra `describe` on BibleCard host-highlights: dropped. New cases are file-scope `it`s.

**Hard (remaining, non-blocking for merge)**

- `docs/testing.md` still wants one optional top-level `describe`. New grouping describes remain in `verse.test.tsx` and `verse-of-the-day.test.tsx`.

**Judgement**

- Duplicated Code: omit-vs-empty paint is restated on BibleCard and VerseOfTheDay.
- Middle Man: VerseOfTheDay `clipHighlightsToPassage` still wraps `filterHighlightsForPassage`.
- Data Clump: `requestedUsfm` and `renderedPassageId` travel through both paint-scope helpers.

## Spec

**(a) Missing / partial:** none.

**(b) Scope creep** (still present; not ticket blockers)

- Dual-prop: `highlights` still wins over `highlightedVerses`. The reader never passes both.
- `chapterScopeForHighlightPaint` still returns `null` on a same-chapter load, so paint goes blank. `2d7206d` keeps paint while a *different* chapter loads. The same-chapter blank is unchanged.

**(c) Implemented but wrong:** none.

---

Standards: 2 remaining hard (extra describes), 3 judgement (worst: extra `describe` in edited files). Spec: 0 missing, 2 scope-creep (worst: same-chapter blank paint). Prior blocking mock leak is closed.

PR #349, opened 2026-08-19
feat(ui): prototype shared shadow overlay for portals (YPE-5138)

abharms:
## Summary

- add one shared, styled overlay shadow root for portalled SDK content
- route internal Radix Popover and Dialog portals to that root when rendered under ShadowRootHost
- prove the architecture with BibleVersionPicker and SignInDialog Storybook browser integration tests
- constrain popovers to Radix's available viewport height and namespace the picker spacing token
- preserve the existing document.body portal behavior outside shadow isolation
- record findings and remaining accessibility risks in ADR 0005

## Scope

This is an architecture spike, not a BibleVersionPicker rollout. The public BibleVersionPicker export is unchanged; the ShadowRootHost wrapper exists only in the spike stories.

## Findings

The Chromium proof covers host-CSS isolation, ancestor and viewport clipping, internal scrolling, positioning, focus, Escape and outside-click dismissal, shared-overlay reuse, and actual SignInDialog composition. Cross-root aria-controls resolution and real VoiceOver/NVDA behavior remain explicit follow-up gates.

## Verification

- UI unit suite: 408 tests passed
- Storybook Chromium integration: 11 tests passed
- UI build and style verification passed
- UI typecheck and lint passed
- git diff --check passed

The root pnpm test command could not complete locally because the core test setup requires YVP_API_HOST, which was not available. The complete UI suite passed independently.

### DIFF
--- a/.changeset/calm-shadows-float.md
+++ b/.changeset/calm-shadows-float.md
@@ -1,0 +1,5 @@
+---
+'@youversion/platform-react-ui': patch
+---
+
+Route isolated Radix popovers and dialogs through a shared styled shadow overlay, and constrain popovers to the available viewport height.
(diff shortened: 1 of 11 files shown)


abharms:
Superseded by #350, which uses a same-repository branch so CI can access the required repository secrets.

PR #360, opened 2026-08-25
feat(hooks): serve cached reads instantly via TanStack Query

Dustin-Kelley:
## What

Replaces the hand-rolled fetch plumbing inside `useApiData` with TanStack Query, so all 14 data hooks (`useChapter`, `useVerses`, `useVersions`, `useHighlights`, …) cache reads: revisiting previously fetched content renders instantly from an in-memory cache and revalidates in the background, instead of blanking and refetching.

**Public hook contracts are unchanged** — every hook still returns `{ data, loading, error, refetch }`, and consumers add no providers and install no new peers. The one surface addition: content hooks now accept an optional `keepPreviousData` flag (default `true`) to opt out of holding the previous key's data on screen across a key change.

## How

- `@tanstack/react-query` is a **direct dependency** of the hooks package, exact-pinned at `5.101.4` (clears the 3-day `minimumReleaseAge` supply-chain window). The ui package gets it as a test-only devDep.
- A **private `QueryClient`** lives inside `YouVersionProvider`; no TQ surface is exported. Its defaults live in `internal/queryClientDefaults.ts`, shared with the test providers: `retry: false` (hooks surface the first failure instead of delaying it behind three retries), `networkMode: 'always'` (see below), `refetchOnWindowFocus: false` (see below), and a `refetchOnReconnect` predicate that refetches only errored reads (see below).
- Query keys derive from **provider config** (`useQueryKeyBase`: appKey, apiHost, installationId, headers key, active Bible version filter) rather than client instances — client identity is a pure function of config, and keys must be serializable. Each hook appends its name plus params. `useQueryKeyBase` serializes `additionalHeaders` itself, because `YouVersionContext` is public: a host can render the context directly and set headers without any precomputed key, and those headers reach the request. The provider's precomputed `additionalHeadersKey` field was dropped as unused.
- `useHighlights` adds a **per-user segment** (`useUserScope`), so one account can never see another account's cached highlights. The scope is the `userId`; `'anon'` when nobody is or can be signed in; or `null` when no safe key exists — auth still loading, a signed-in profile whose optional `userId` has not resolved, or an access token set while the auth context reports no user (the request layer and the context disagree about identity, so even `'anon'` is unsafe). A `null` scope **disables the query**, because two unidentified accounts would otherwise share one cache entry. Disabled hooks additionally mask `data`/`error` to `null`. Account-scoped hooks also pass `keepPreviousData: false`, so an account switch drops to `null` rather than holding the previous account's data on screen.
- The Bible reader's write gate matches that read gate: a profile counts as signed in only when auth has finished loading **and** the profile carries a `userId`. Otherwise a highlight POST could succeed against an account whose reads are disabled — the overlay paints and the highlight is gone on the next load. The highlights machine also tracks the account id and resets its write state (optimistic overlay, reconcile map, write-intent tokens, queue) whenever that id changes, covering hosts that swap `userInfo` directly from one signed-in user to another without a sign-out.
- `refetch` maps to **exact query invalidation**, so the highlights xstate machine keeps owning writes and its post-write refresh works across remounts.
- Cache is **memory-only** (TQ defaults: staleTime 0, gcTime 5 min); nothing persists.
- Removed the referential-stability workaround in `use-bible-reader-highlights` — TQ structural sharing keeps unchanged data reference-stable, so a plain `useMemo` suffices.
- Decision record: `docs/adr/0006-tanstack-query-read-layer.md`.

### Request-traffic and offline behavior

Two defaults exist to keep this layer's network behavior at or below the plumbing it replaced:

- **`networkMode: 'always'`** — TQ pauses fetches by default when the browser reports it is offline. A paused first load stays `loading: true` forever and never settles; a paused fetch after a key change shows the previous key's data as if it belonged to the new key. Hooks now always attempt the request and report transport failures through `error`, as before. TQ infers `refetchOnReconnect` from `networkMode`, which would leave it off entirely; it is instead a predicate that refetches **only errored reads** — a read that failed while the connection was down recovers on its own when it returns, while settled reads keep their data instead of revalidating all at once on every reconnect.
- **`refetchOnWindowFocus: false`** — focus revalidation was the only net-new request traffic the migration introduced, and a reader holds several queries at once. Most reads are Bible content, which does not change while the reader is open. Mounting, a key change, and `refetch()` all still revalidate. Chosen over a blanket `staleTime`, which would also suppress revalidation on mount and on re-enable, weakening the cache contract below the pre-TQ layer's freshness.

## Main files to review

The diff is wide but shallow — twelve of the hook files are a thin call-site swap. The actual logic is in these files:

| File | What to look at |
|---|---|
| [`packages/hooks/src/useApiData.ts`](https://github.com/youversion/platform-sdk-react/blob/dk/tanstack-query-implementation/packages/hooks/src/useApiData.ts) | **The core rewrite.** The whole fetch layer, replaced. Check the `{ data, loading, error, refetch }` mapping — particularly what `loading` covers (first load of a key, the placeholder window after a key change, and a fetch retrying an errored read, but *not* background revalidation of settled data), the `null` masking when `enabled` is false, and why `refetch` is `invalidateQueries({ exact: true })` rather than `query.refetch()`. |
| [`packages/hooks/src/internal/queryClientDefaults.ts`](https://github.com/youversion/platform-sdk-react/blob/dk/tanstack-query-implementation/packages/hooks/src/internal/queryClientDefaults.ts) | **Every network-behavior policy in one place.** `YouVersionProvider` builds its private `QueryClient` from this object: `retry`, `networkMode`, `refetchOnWindowFocus`, and the `refetchOnReconnect` predicate. Each one is a deliberate deviation from TQ's defaults with a comment explaining the trade-off. This is where request traffic is decided. |
| [`packages/hooks/src/useOrganizations.ts`](https://github.com/youversion/platform-sdk-react/blob/dk/tanstack-query-implementation/packages/hooks/src/useOrganizations.ts) | **The one hook that reaches TanStack Query directly.** `useApiData` wraps a single `useQuery` and cannot express a batch, so this calls `useQueries` with one query per id, keyed exactly as `useOrganization` keys it. Check the id-list identity string and the stable `combine` identity — both exist to keep the returned `Map` referentially stable. |
| [`packages/hooks/src/internal/useUserScope.ts`](https://github.com/youversion/platform-sdk-react/blob/dk/tanstack-query-implementation/packages/hooks/src/internal/useUserScope.ts) | **New file — the account-isolation rule.** Three states: a `userId`, `'anon'`, and `null`. The `null` cases are the load-bearing ones — auth loading, a missing/empty id, or a token that contradicts the signed-out context — each disables the query, because two unidentified accounts would otherwise share a single cache entry. |
| [`packages/ui/src/components/use-bible-reader-highlights.ts`](https://github.com/youversion/platform-sdk-react/blob/dk/tanstack-query-implementation/packages/ui/src/components/use-bible-reader-highlights.ts) | **The write gate.** Aligns "can write" with "can read" — signed in requires a resolved (`!isLoading`) profile with a `userId` — and feeds the account id into the highlights machine. Also drops the old referential-stability workaround now that TQ structural sharing covers it. |
| [`packages/ui/src/components/bible-reader-highlights-machine.ts`](https://github.com/youversion/platform-sdk-react/blob/dk/tanstack-query-implementation/packages/ui/src/components/bible-reader-highlights-machine.ts) | **The account-change reset.** `AUTH_CHANGED` carries the account id; `resetWriteStateOnAccountChange` clears the overlay, reconcile map, write-intent tokens, and queue on sign-out or on an id change. It runs before `assignAuth`, while context still holds the previous id, and the cleared write-intent tokens defuse the previous account's in-flight writes. |
| [`packages/hooks/src/useHighlights.ts`](https://github.com/youversion/platform-sdk-react/blob/dk/tanstack-query-implementation/packages/hooks/src/useHighlights.ts) | **The only hook with real read policy of its own.** The `enabled` gate and the hardcoded `keepPreviousData: false` (an account-safety invariant, not caller-overridable). Everything else it does is shared with the other thirteen hooks. |

Lower priority, mechanical: the key-derivation helpers (`useQueryKeyBase.ts`, `versionFilterKey.ts`, `additionalHeadersKey.ts`) are pure serialization, and the twelve remaining `use*.ts` hooks repeat the same call-site swap plus the `keepPreviousData` pass-through.

## Verification

- core 431/431; hooks 318/318; ui 517/517 (incl. all `use-bible-reader-highlights` integration suites); typecheck, build, and `pnpm lint` clean. Re-run green after merging `main`.
- New tests: cache-on-revisit, A/B/A user-scoping, hold-off while the account is unidentified, the anonymous scope withheld while a token contradicts it, memory-only behavior, non-Error normalization, re-enable-from-cache, keep-previous-data across a key change (and `keepPreviousData: false` dropping it), offline reads still settling, reconnect refetching only errored reads, `loading` shown while a refetch retries an errored read, no refetch on window focus (and `refetch()` still revalidating), sibling providers with different `additionalHeaders` not sharing a cache entry, the reader treating an unresolved or id-less profile as signed out, an account swap without a sign-out clearing optimistic highlights (machine and adapter), SSR smoke (`@vitest-environment node`).
- Browser-verified in `examples/vite-react`: with chapter responses delayed 3 s, a revisited chapter still rendered instantly from cache (~30 ms) and issued a background revalidation GET.

## Review follow-ups

Feedback from review threads, each landed as its own commit:

- b330e1c — `useUserScope` returns `null` (fetch off) while a stored access token contradicts the signed-out auth context.
- 48e1336 / c148fb2 — reconnect refetches only errored reads instead of every mounted query.
- 9f29560 — `loading` also covers a refetch that retries an errored read.
- 587599e — dropped the unused precomputed `additionalHeadersKey` context field.
- 44533d9 — highlight writes gated while auth is still loading.
- 173b4d9 — content hooks forward the `keepPreviousData` option.
- 5f2d407 — the highlights machine resets its write state when the account id changes (the direct A→B swap without a sign-out).
- 8608a98 — the render masks the one-paint window before the `AUTH_CHANGED` effect delivers that reset, mirroring the existing scope-change mask.
- d37747e — every read keyed by the active Bible version filter; 41f83f9 — organization cache shared across mounts and hooks; 1166932 — empty user id treated as unidentified.

## Reviewer notes

- **Behavioral deltas** (accepted in ADR 0006): revisits show cached data with background revalidation instead of a loading blank; `refetch` no longer flips `loading` when data is already on screen; returning to a tab no longer refetches. `useHighlights` does not fetch until the signed-in account is identified.
- `loading` covers the first load of a key (`isPending`), the placeholder window after a key change (`isPlaceholderData && isFetching`), and a fetch retrying an errored read (`isError && isFetching`); a background revalidation of settled data leaves it `false`.
- `useOrganizations` is the one hook that bypasses `useApiData` — not because it is un-migrated, but because `useApiData` wraps a single `useQuery` and a batch needs one query per id. It calls `useQueries` directly, keyed exactly as `useOrganization` keys it, so the two share cache entries; no TanStack Query type reaches its public surface. Called out in `packages/hooks/AGENTS.md`.
- `useVOTD.test.tsx` semantics changed: cache identity keys on provider config, so a config change is what invalidates previously fetched data.
- Test env fix, split out as its own commit: Node 26 ships a flagless `localStorage` accessor that shadows jsdom's; both packages' test setup now graft a working Storage trio from a private JSDOM window (fixes parallel-run flakes).
- `react-dom` joins the hooks devDeps because the new SSR smoke test imports `react-dom/server`.
- Remaining human check: a signed-in account-switch smoke test in `examples/vite-react` (needs real credentials).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<h3>Greptile Summary</h3>

The PR migrates the hooks read layer to a private TanStack Query cache while preserving the public hook result contract.
- Adds provider-scoped query clients, configuration- and user-scoped cache keys, explicit network/refetch defaults, and batched organization queries.
- Resets account-owned highlight write state on direct account changes and masks stale optimistic overlays during the pre-effect render.
- Adds caching, account-isolation, SSR, reconnect, and highlight-account-transition coverage.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/hooks/src/useApiData.ts | Replaces local request state with TanStack Query while retaining the public data, loading, error, and refetch shape. |
| packages/hooks/src/internal/useUserScope.ts | Derives fail-closed account cache scopes and disables authenticated reads when identity is unresolved. |
| packages/hooks/src/useHighlights.ts | Keys highlight reads by account and prevents previous-account data from surviving key transitions. |
| packages/ui/src/components/bible-reader-highlights-machine.ts | Resets account-owned write state and disowns late completions so old-account permission failures cannot be resumed under the current account. |
| packages/ui/src/components/use-bible-reader-highlights.ts | Masks optimistic state until the machine account matches the currently rendered account. |
| packages/hooks/src/useOrganizations.ts | Moves organization lookups into shared per-ID queries while preserving a stable combined Map. |

<sub>Reviews (12): Last reviewed commit: ["Merge remote-tracking branch &#39;origin/mai..."](https://github.com/youversion/platform-sdk-react/commit/9ca682887af029b47cb59607067d84688d9a2664) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57070211)</sub>

- Knowledge Base — [Hook data-fetching lifecycle](https://app.greptile.com/youversion/-/custom-context/knowledge-base/youversion/platform-sdk-react/-/docs/hooks-data-fetching.md)
- Knowledge Base — [React hooks integration](https://app.greptile.com/youversion/-/custom-context/knowledge-base/youversion/platform-sdk-react/-/docs/react-hooks.md)
- Knowledge Base — [Scripture highlighting interactions](https://app.greptile.com/youversion/-/custom-context/knowledge-base/youversion/platform-sdk-react/-/docs/ui-highlights.md)
</details>


packages/hooks/src/internal/useUserScope.ts
```diff
@@ -0,0 +1,45 @@
...
+  const auth = useContext(YouVersionAuthContext);
+  // No auth context: the host never signs anyone in, so there is one scope.
+  if (!auth) return 'anon';
+  // Auth is resolving. The scope is not known yet.
+  if (auth.isLoading) return null;
+  // Auth resolved with no user.
+  if (!auth.userInfo) return 'anon';
+  // Signed in. A missing id leaves the account unidentified, so there is no
+  // safe key for it.
+  return auth.userInfo.userId ?? null;
```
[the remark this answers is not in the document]
Dustin-Kelley: **note:** Fixed in `1166932`.

Confirmed: `YouVersionUserInfoJSONSchema` has `id: z.string().optional()`, so `''` parses, and `useHighlights` enables on `scope !== null` — every profile carrying an empty id keyed to one cache entry, and a switch between two of them showed the first account's highlights with no refetch.

`useUserScope` now returns `null` for an empty id, the same handling a missing id already got, so the query stays disabled until a non-empty identifier arrives. The existing unidentified-account test walks the empty-id step alongside resolving-auth and missing-id.

*Posted by Claude Code, an AI agent, on behalf of Dustin.*

packages/hooks/src/internal/useQueryKeyBase.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+export function useQueryKeyBase(): readonly unknown[] {
+  const context = useContext(YouVersionContext);
+  const additionalHeaders = context?.additionalHeaders;
+  const additionalHeadersKey = useMemo(
+    () => serializeAdditionalHeaders(additionalHeaders),
+    [additionalHeaders],
+  );
+
+  return [context?.appKey, context?.apiHost, context?.installationId, additionalHeadersKey];
```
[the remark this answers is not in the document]
Dustin-Kelley: **note:** Fixed in `d37747e`.

Agreed — the filter decides what a read may return, so it belongs in the key next to the app key, host, installation id, and headers. It lives on `YouVersionPlatformConfiguration` (the provider writes it from props during render), which is why it never reached the key.

`useQueryKeyBase` now appends `serializeVersionFilters()`. Ids and tags are sorted so two equivalent lists agree, and an unset list serializes as `null` so "no filter" stays distinct from "an empty filter list". Two tests added: a tightened filter gets its own cache entry (verified failing without the change), and list order alone does not split entries.

*Posted by Claude Code, an AI agent, on behalf of Dustin.*

packages/hooks/src/useApiData.ts
```diff
@@ -1,95 +1,106 @@
...
-  // actually triggers the fetch.
-  useEffect(() => {
-    fetchData();
-    return () => {
-      // Invalidate any in-flight request (effect- or refetch-initiated) when
-      // the deps change or the component unmounts.
-      requestSeqRef.current++;
-    };
-  }, [...deps, enabled]);
+    void queryClient.invalidateQueries({ queryKey: queryKeyRef.current, exact: true });
```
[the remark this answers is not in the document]
Dustin-Kelley: **note (non-blocking):** Not changing this — the proposed fix would be a no-op.

Invalidating user A's key instead of B's only marks A's entry *stale*. The QueryClient leaves `staleTime` at its default `0` and `refetchOnMount` on, so A's entry is already stale: switching back serves the cached highlights and fires a background revalidation either way. The observable behavior is identical with or without the invalidation reaching A.

That cache-first-then-revalidate flash is the deliberate design here, covered by the "should never show one user their predecessor cache, and restore it instantly on switch-back" test in `useHighlights.test.tsx`. Evicting or otherwise suppressing A's entry would reverse it.

Separately, `useApiData` has no way to know which key was live when a *write* started — only the caller holds that. Targeting the write-time key would have to be a change in the highlights machine, not in this layer.

*Posted by Claude Code, an AI agent, on behalf of Dustin.*

Dustin-Kelley:
**note (non-blocking):** Decision on the "refetch targets wrong account" finding — not fixing in this PR, and not treating it as a merge blocker.

Recording this at the top level so it's carried into the next review pass. The finding is [`useApiData.ts:76`](https://github.com/youversion/platform-sdk-react/blob/d37747e9458e2b9a1d9be5aea8e6cc145f6d581a/packages/hooks/src/useApiData.ts#L76), and it's the sole basis for the current Confidence Score of 4/5.

### The mechanism is real

`refetch` is stable and reads `queryKeyRef.current`, a ref reassigned every render. The highlights machine fires its settle-time GET through `servicesRef.current.refetch()` — also a latest-ref — and the actor is spawned once via `useActorRef`, so it survives an auth change rather than re-spawning. If a write for account A settles after the key has moved to B, `invalidateQueries({ exact: true })` lands on B. A's entry is never invalidated. That sequence is accurate as described.

Worth noting the framing is slightly narrower than the actual behavior: it doesn't need to be a *direct* A→B switch. A → signed-out → B misses identically. The invalidation targets whatever key is current at settle time, which is never the write-time key. Only the identity of the wrong target changes.

### Why it's not being fixed here

**1. No observable difference today.** The QueryClient sets neither `staleTime` nor `refetchOnMount`, so both sit at TanStack Query's defaults (`0` and `true`). A's entry is already stale the instant it lands — marking it stale again is a no-op. Switching back to A serves the cached highlights and fires a background revalidation either way. This was already agreed on the inline thread.

**2. The flow requires a deliberate human action inside the write window.** Nothing changes `userInfo` automatically. `YouVersionAuthProvider`'s token refresh runs once in a mount-time effect with `[]` deps — no interval, no `visibilitychange`, no mid-session re-identification anywhere in hooks or core. The only triggers are mount, a `signOut()` click, or a host swapping the prop. So the window is bounded by how fast a person can tap a highlight color and then change accounts.

**3. It's unreachable on the web path.** `signIn()` is a full-page OAuth redirect and the cache is memory-only (no persister in the repo), so every web sign-in discards the entire QueryClient. A's stale entry cannot survive to be seen. The flow is confined to host-controlled auth — `<YouVersionAuthProvider userInfo={...}>` or the `onSignInPress`/`onSignOutPress` native bridges — where the account changes in place with no reload.

Mechanically real, observably inert, and reachable only through a contrived sequence on one auth path. That does not clear the bar for this PR.

### The condition that makes it real

A `staleTime > 0`. At that point switching back to A serves A's entry as fresh, no revalidation fires, and the highlight A just created is missing from their screen until the timer expires. Whoever introduces `staleTime` has to address this in the same change.

The fix does not belong in `useApiData`, which has no way to know which key was live when a *write* began — only the caller holds that. Targeting the write-time key is a change in `bible-reader-highlights-machine.ts`.

### For the next pass

The 4/5 summary reflects a review at `d37747e`, which is now several commits behind. Two fixes have landed since (`refetchOnReconnect`, and migrating `useOrganizations` onto the query layer), plus a merge of `main`. Please re-evaluate against the current tip and treat this finding as accepted-and-deferred rather than outstanding.

*Posted by Claude Code, an AI agent, on behalf of Dustin.*

packages/ui/src/components/use-bible-reader-highlights.ts
```diff
@@ -180,7 +180,13 @@ export function useBibleReaderHighlights({
...
   const authContext = useContext(YouVersionAuthContext);
   const hasAuthProvider = authContext !== null;
-  const isAuthenticated = Boolean(authContext?.userInfo);
+  // A profile with no `userId` is not enough. `useHighlights` keys its cache by
+  // account and does not fetch for an unidentified one, so treating this state
+  // as authenticated would allow writes that no read can ever return: the POST
+  // succeeds, the overlay paints, and the highlight is gone on the next load.
+  // A host that supplies `userInfo` itself can produce this state, because
+  // `userId` is optional on the profile.
+  const isAuthenticated = Boolean(authContext?.userInfo?.userId);
```
cameronapak: **question:** Write gate does not wait for `auth.isLoading`.

Reads use `useUserScope()`. That function returns `null` while `auth.isLoading` is true. `useHighlights` then does not fetch.

Writes use `Boolean(authContext?.userInfo?.userId)` only.

The default `YouVersionAuthProvider` starts load with `userInfo` as `null`. That path is safe.

A host that keeps a leftover `userId` on the profile while `isLoading` is true can POST while the GET stays off.

Should the write gate also require `!authContext.isLoading`?

ADR 0006 names the no-`userId` case. It does not name the loading case.

---

packages/hooks/src/useChapter.ts
```diff
@@ -13,15 +14,16 @@ export function useChapter(
...
   const bibleClient = useBibleClient();
+  const keyBase = useQueryKeyBase();
 
   const {
     data: chapterData,
     loading,
     error,
     refetch,
   } = useApiData<BibleChapter>(
+    [...keyBase, 'chapter', versionId, book, chapter],
```
cameronapak, line 25: **nit:** `keepPreviousData` on the public options does nothing here.

`UseApiDataOptions` includes `keepPreviousData`. This wrapper only forwards `enabled`.

A caller who passes `keepPreviousData: false` still holds previous data.

If the public type is the contract, forward the field. If not, drop it from the wrapper type.

The other fetch wrappers do the same drop.

---

packages/hooks/src/context/YouVersionContext.tsx
```diff
@@ -13,6 +13,16 @@ export type YouVersionContextData = {
   theme?: 'light' | 'dark';
   authEnabled?: boolean;
   additionalHeaders?: Record<string, string>;
+  /**
+   * @internal
+   *
+   * A JSON string of `additionalHeaders` with the keys in sorted order.
+   * When `additionalHeaders` is not set, the value is `null`.
+   * `YouVersionProvider` builds this string once.
+   * The memo for the API client and the `queryKey` both use this string.
```
cameronapak: **nit:** This comment says the query key uses `additionalHeadersKey`.

`useQueryKeyBase` serializes `context.additionalHeaders` itself. That is the spec.

A host that sets `additionalHeaders` and leaves `additionalHeadersKey` unset still gets a correct key.

Please update the comment so it matches the hook.

---

cameronapak (commented):
## Spec

No must-fix. The spike matches the PR body and ADR 0006.

YPE-4153 acceptance criteria did not load from Jira. I judged the work against the ticket title, the PR body, and ADR 0006.

Counts: 0 must-fix. 1 question. 3 nits.
Worst issue: the write gate does not wait for `auth.isLoading`.

The default `YouVersionAuthProvider` starts load with `userInfo` as `null`. That path is safe.

A host that keeps a leftover `userId` while `isLoading` is true can POST while the GET stays off.

## Standards

No must-fix.

Counts: 0 must-fix. 0 questions. 3 nits.
Worst issue: public wrappers accept `keepPreviousData` and then drop it.

`useChapter` and the other fetch wrappers only forward `enabled`. A caller who passes `keepPreviousData: false` still holds previous data.

`TestQueryClientProvider` only sets `retry: false`. Production also sets `networkMode`, `refetchOnWindowFocus`, and `refetchOnReconnect`.

`staleTime` and `gcTime` stay at TanStack Query defaults. The ADR names those values. The client does not set them.

```
useApiData
  queryKey: [...useQueryKeyBase(), hookName, params]
  user hooks also add useUserScope()
YouVersionProvider
  private QueryClient (useState, not a prop)
```

packages/ui/src/components/use-bible-reader-highlights.ts
Dustin-Kelley: **note:** Fixed in 44533d9. `isAuthenticated` now requires `!authContext.isLoading` alongside the `userId` check, so the write gate is withheld in the same window where `useUserScope` withholds the read. The tradeoff: a color tap during the brief loading window is treated as signed out and opens the sign-in dialog instead of writing. Added a test covering a profile with a valid `userId` but `isLoading: true` — the fetch stays disabled and a tap produces no POST.

— Claude, an AI agent posting on behalf of Dustin.

packages/hooks/src/useChapter.ts
Dustin-Kelley: **note:** Fixed in 173b4d9. All thirteen fetch wrappers now forward `keepPreviousData` to `useApiData`, so the public options type is honored (defaults unchanged). `useHighlights` keeps its hardcoded `false` — that one is an account-safety invariant, not a drop. Added a test asserting `useChapter(…, { keepPreviousData: false })` drops to `null` while a chapter change fetches.

— Claude, an AI agent posting on behalf of Dustin.

packages/hooks/src/context/YouVersionContext.tsx
Dustin-Kelley: **note:** Already fixed in 587599e — the `additionalHeadersKey` field and its docstring were removed from `YouVersionContextData` entirely. The key segment is derived inside `useQueryKeyBase` from `context.additionalHeaders`, so the stale claim no longer exists.

— Claude, an AI agent posting on behalf of Dustin.

packages/hooks/src/useHighlights.ts
```diff
@@ -35,18 +37,30 @@ export function useHighlights(
...
+  //
+  // A `null` `userScope` means the account is not known: auth is still loading,
+  // or the signed-in user has no id. Highlights belong to one account, so this
+  // hook does not fetch until the scope is known. Two unidentified accounts
+  // would otherwise write to the same cache entry.
   const { data, loading, error, refetch } = useApiData<Collection<Highlight>>(
+    [...keyBase, 'highlights', userScope ?? 'unresolved', options.version_id, options.passage_id],
     () => highlightsClient.getHighlights(options),
-    [highlightsClient, options.version_id, options.passage_id],
     {
```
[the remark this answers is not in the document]
Dustin-Kelley: **note (non-blocking):** Not changing this here — it is the direct authenticated→authenticated switch scenario already deferred in the PR discussion. The machine clears its overlay on sign-out (`resetWriteStateIfSignedOut`) and on scope change, and the web OAuth flow always passes through sign-out plus a full-page redirect. A direct A→B swap with no intermediate signed-out state is only reachable via host-controlled `userInfo`, which is the deferred case. If we close it later, the fix is to thread the user id into `AUTH_CHANGED` and reset write state when it changes.

— Claude, an AI agent posting on behalf of Dustin.

Dustin-Kelley:
@greptile, please see my comment https://github.com/youversion/platform-sdk-react/pull/360#discussion_r3866857023 and review again

packages/ui/src/components/use-bible-reader-highlights.ts
```diff
@@ -235,38 +247,35 @@ export function useBibleReaderHighlights({
...
+      userId,
     },
   });
 
   // ── Feed React-owned inputs to the machine ──────────────────────────────────
   useEffect(() => {
-    actorRef.send({ type: 'AUTH_CHANGED', flagOn, hasAuthProvider, isAuthenticated });
-  }, [actorRef, flagOn, hasAuthProvider, isAuthenticated]);
+    actorRef.send({ type: 'AUTH_CHANGED', flagOn, hasAuthProvider, isAuthenticated, userId });
+  }, [actorRef, flagOn, hasAuthProvider, isAuthenticated, userId]);
```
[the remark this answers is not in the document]
Dustin-Kelley: **issue (resolved):** Fixed in 8608a98. The persistent half of this was already closed in 5f2d407 (the machine resets its write state when the account id changes), and this comment correctly narrowed to the remaining one-paint window: `AUTH_CHANGED` is sent from a `useEffect`, which runs after the swap render commits.

The render now masks that window the same way it already masks a synchronous scope change: it compares the machine's account id (`state.context.userId`) with the render's and returns server truth alone while they differ, so the previous account's overlay is never merged into the new account's view, not even for one frame. Covered by a test that records every render — including the pre-effect swap render — and asserts none of them contains the previous account's overlay (red-checked against the unmasked code: exactly that test fails).

— Claude, an AI agent posting on behalf of Dustin

packages/ui/src/components/bible-reader-highlights-machine.ts
```diff
@@ -415,12 +424,20 @@ export const bibleReaderHighlightsMachine = setup({
...
 
-    /** Sign-out / going disabled must not leave a stale overlay to resurface later. */
-    resetWriteStateIfSignedOut: assign(({ event }) => {
-      if (event.type !== 'AUTH_CHANGED' || event.isAuthenticated) return {};
+    /**
+     * Clears the write state on sign-out or on an account change, so the next
+     * account does not see optimistic entries from the previous one. The
+     * cleared `writeIntent` tokens also defuse in-flight writes: a write that
+     * settles later fails the per-verse ownership check, the same as after a
+     * scope change.
```
cameronapak, line 436: **question:** Cleared tokens do not defuse a late 401 or 403.

This reset clears `writeIntent`. `settleWrite` then skips overlay and reconcile updates when the token is gone.

The 401/403 path still calls `appendPendingHighlight` and raises `PERMISSION_LOST`. There is no token check on that path.

A write from account A that settles after an A-to-B swap can open the permission dialog on account B.

The comment says this is the same as a scope change. That match holds. The spec line about defusing in-flight writes does not, if it means no prior-account side effects.

Is the host A-to-B swap still deferred, or should this path also check the token?

---

packages/hooks/src/internal/queryClientDefaults.ts
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Query defaults for every `QueryClient` in this repository.
+ *
+ * `YouVersionProvider` builds its private client from this object. Test
+ * providers do the same, so tests run with the fetch semantics of the app.
+ * The `TestQueryClientProvider` in `packages/ui` cannot import this module —
+ * the package exports only its root — so it holds a mirrored copy. A change
+ * here must also change that copy.
+ *
+ * The reasons for each option are in `docs/adr/0006`.
```
cameronapak, line 12: **question:** This comment says the reasons for each option live in ADR 0006.

`refetchOnReconnect` is a predicate that refetches only errored reads. That policy is in the PR body. It is not in the ADR.

`loading` also covers `isError && isFetching`. The ADR still names only `isPending` and the placeholder window.

`useUserScope` now returns `null` when a token contradicts an empty user. The ADR still names only loading and a missing `userId`.

Please update the ADR, or stop pointing here as the full reason list.

---

cameronapak (commented):
## Spec

Prior findings are fixed in code. Write gate waits for `!isLoading`. Wrappers forward `keepPreviousData`. `additionalHeadersKey` is gone.

No must-fix. Counts: 0 must-fix. 1 question. 2 nits.
Worst issue: a late 401 or 403 from account A can still open the permission dialog on account B.

`resetWriteStateOnAccountChange` clears overlay, reconcile, tokens, and the queue. `settleWrite` then skips overlay updates when the token is gone. The 401/403 path still calls `appendPendingHighlight` and raises `PERMISSION_LOST` with no token check.

If "defuse" means no side effects from the previous account, this is incomplete. If it means "same as a scope change," it matches. The web OAuth path goes through sign-out, so this is a host A-to-B swap.

Nits: ADR 0006 does not name error-retry loading, token-contradicts-anon, or error-only reconnect. `useHighlights` still documents only loading and a missing id.

## Standards

No must-fix. Counts: 0 must-fix. 1 question. 1 nit.
Worst issue: ADR 0006 is behind this delta, and `queryClientDefaults.ts` says the reasons live there.

The new reconnect predicate and the `isError && isFetching` loading case are in the code and the PR body. They are not in the ADR.

Nit: the title still includes `(YPE-4153)`. Release-hardening Decision 1 keeps tickets in the body and the branch.

packages/hooks/src/internal/queryClientDefaults.ts
cameronapak: **issue:** Pre-flight is DO NOT SHIP until ADR 0006 names these two rules.

1. `refetchOnReconnect` is a predicate that refetches only errored reads. See `queryClientDefaults.ts` line 44.
2. `loading` is also true when `isError && isFetching`. See `useApiData.ts` lines 108-112.

ADR Consequences still say `loading` is only `isPending` and the placeholder window. The ADR does not name the reconnect predicate.

The PR body already holds both contracts. Please add them to `docs/adr/0006-tanstack-query-read-layer.md`.

Not blocking: hooks minor stays. Host A-to-B late 401 stays deferred.

---

packages/hooks/src/context/YouVersionProvider.tsx
```diff
@@ -179,17 +192,23 @@ function YouVersionProviderInner(
...
       <YouVersionContext.Provider value={contextValue}>
-        <Suspense>
-          <AuthProvider
-            config={{ appKey, apiHost, redirectUri: props.authRedirectUrl }}
-            userInfo={props.userInfo}
-          >
-            {children}
-          </AuthProvider>
-        </Suspense>
+        <QueryClientProvider client={queryClient}>
```
bmanquen: **issue (blocking):** Nesting `QueryClientProvider` (here and at line 211) shadows the host's client whenever their `@tanstack/react-query` dedupes to our pinned `5.101.4` — their components inside `YouVersionProvider` then run against our cache with our defaults (`retry: false`, `refetchOnWindowFocus: false`), and their own `invalidateQueries` / `dehydrate` no longer reach those queries. Nothing throws; it just silently stops working. Whether it happens depends on the host's lockfile, so it's non-deterministic across consumers.

ADR 0006 accepts the duplicate-instance cost ("a second client and cache copy — bounded, memory-only") but not this deduped case.

Since no consumer ever needs our client, we can drop `QueryClientProvider` entirely and pass the client explicitly — v5 takes it as the second arg to `useQuery` / `useQueries` / `useQueryClient`. Carry it on a private internal context instead. Touches `useApiData.ts:52,54`, `useOrganizations.ts:65`, and both `TestQueryClientProvider` seams.

packages/hooks/src/internal/useUserScope.ts
```diff
@@ -0,0 +1,53 @@
...
+ * This hook reads `YouVersionAuthContext` directly. It does not call
+ * `useYouVersionAuthContext()`, because that call throws when auth is off,
+ * and data hooks must work without auth. No auth context means that the host
+ * owns identity, so one scope covers any token the host sets. The React
+ * Native Expo SDK also puts `userInfo` in this context.
+ */
+export function useUserScope(): string | null {
+  const auth = useContext(YouVersionAuthContext);
+  // No auth context: the host owns identity, and one scope covers it.
+  if (!auth) return 'anon';
```
bmanquen: **issue (blocking):** `if (!auth) return 'anon'` is justified as "the host owns identity, so one scope covers any token the host sets" — but that holds only if the host's identity never changes. With `includeAuth: false` and a host-set `YouVersionPlatformConfiguration.accessToken`, swapping tokens between accounts puts both under the same `'anon'` highlights entry, and default `gcTime` keeps it live for 5 minutes. That's the exact leak the `null` scope exists to prevent — the line below already returns `null` for this disagreement when an auth context happens to exist.

The RN Expo path is safe (it supplies `userInfo` through the context), so this is specifically `includeAuth: false` + manual token. If that's a supported way to use `useHighlights`, the no-context branch needs the same token check; if it isn't, the docstring should say so.

bmanquen (commented):
Two blocking findings from a review of `main...HEAD`. Posted by Claude Code, an AI agent, on behalf of Brenden.

packages/hooks/src/internal/useQueryKeyBase.test.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+    await waitFor(() => {
+      expect(getVOTD).toHaveBeenCalledTimes(1);
+    });
+    expect(queryClient.getQueryCache().getAll()).toHaveLength(1);
+  });
+});
+
+describe('useQueryKeyBase — version filters', () => {
+  afterEach(() => {
```
bmanquen: **issue (non-blocking):** Two `docs/testing.md` rules in one spot. "No `beforeEach`/`afterEach`; inline setup or call factories" — this `afterEach` resets the `YouVersionPlatformConfiguration.permitted*` statics, which also trips "No shared mutable state across cases." And "one optional top-level `describe` for the module" — this is the file's second (the first is line 35).

A helper that sets and restores the statics inline per case satisfies both.

packages/hooks/src/useApiData.test.tsx
```diff
@@ -104,45 +222,179 @@ describe('useApiData — stale responses (latest wins)', () => {
...
     });
     expect(result.current.error).toBeNull();
     expect(result.current.data).toBe('JHN.4 data');
   });
 });
 
+describe('useApiData — offline', () => {
+  // `onlineManager` is module-level state shared by every QueryClient in the
+  // process, so every test here hands it back online.
+  afterEach(() => {
```
bmanquen: **issue (non-blocking):** Same `docs/testing.md` `afterEach` rule as the other file — this one restores `onlineManager`, and line 539 restores the focus manager. Both are undoing module-singleton mutation, which the doc's "No shared mutable state across cases" also covers.

Separately, the file is now at six top-level `describe`s against the doc's "one optional top-level `describe` for the module." Three predate this branch, so this is aggravation rather than a fresh breach — worth a cleanup pass while the file is already open.

packages/ui/src/test/hook-overrides.tsx
```diff
@@ -1,7 +1,38 @@
...
+ *
+ * The defaults mirror `queryClientDefaultOptions` in
+ * `packages/hooks/src/internal/queryClientDefaults.ts` — the hooks package
+ * exports only its root, so this copy cannot import that module. A change
+ * there must also change this copy.
+ */
+export function TestQueryClientProvider({ children }: { children: ReactNode }): ReactElement {
+  const [queryClient] = useState(
+    () =>
+      new QueryClient({
```
bmanquen: **issue (non-blocking):** This hand-mirrors `queryClientDefaultOptions`, and the docstring above says so outright — "A change there must also change this copy," with the reciprocal warning at `queryClientDefaults.ts:9`. That's Shotgun Surgery encoded as a comment: the two copies drift silently, and tests then run under different fetch semantics than the app.

A `./test-utils` subpath in the hooks package `exports` map would let this import the real object. `CONTRIBUTING.md` already covers the mechanics.

packages/ui/src/test/setup.ts
```diff
@@ -3,6 +3,34 @@ import { afterEach } from 'vitest';
 import { cleanup } from '@testing-library/react';
 import '@testing-library/jest-dom/vitest';
 
+// Node >= 22 defines its own experimental `localStorage` global that evaluates
```
bmanquen: **issue (non-blocking):** Lines 1-33 here are byte-identical to `packages/hooks/src/test/setup.ts`, comment included — only ui's trailing `scrollIntoView` stub differs. The Node/jsdom Web Storage graft is subtle enough that a fix applied to one copy and not the other would be hard to spot.

Same remedy as the `TestQueryClientProvider` duplication: one shared test-utils entry point.

packages/hooks/src/useHighlights.ts
```diff
@@ -35,18 +37,30 @@ export function useHighlights(
...
+  // Callers often pass a new `{ version_id, passage_id }` on each render.
+  // A new object does not start a new fetch.
+  // If `GetHighlightsOptions` gets a new field that changes the result, that field must go in the `queryKey` too.
+  //
+  // A `null` `userScope` means the account is not known: auth is still loading,
+  // or the signed-in user has no id. Highlights belong to one account, so this
+  // hook does not fetch until the scope is known. Two unidentified accounts
+  // would otherwise write to the same cache entry.
   const { data, loading, error, refetch } = useApiData<Collection<Highlight>>(
+    [...keyBase, 'highlights', userScope ?? 'unresolved', options.version_id, options.passage_id],
```
bmanquen: **suggestion (non-blocking):** `userScope ?? 'unresolved'` mints a fourth magic string that can never reach the network — line 59 disables the query for exactly the `null` case this fallback covers. It's a dead cache key, and a reader has to check `enabled` to know that.

`useUserScope(): string | null` is carrying four domain states (real id, `'anon'`, unidentified, and now this) in a nullable string. Either give it a small `UserScope` type, or drop the segment entirely when the scope is `null` — the query can't run either way.

packages/hooks/src/internal/queryClientDefaults.ts
```diff
@@ -0,0 +1,46 @@
...
+    refetchOnWindowFocus: false,
+    // A read that failed fetches again when the network returns.
+    // A settled read keeps its data and does not refetch, so a
+    // reconnect does not revalidate every mounted query at once.
+    // TanStack Query infers this option from `networkMode`, and
+    // `always` makes that inference `false`.
+    // With `retry: false` and `refetchOnWindowFocus: false`, an
+    // errored read has no other way back — only a remount or a page
+    // reload would fetch again.
+    refetchOnReconnect: (query) => query.state.status === 'error',
```
bmanquen, line 44: **issue (non-blocking):** `refetchOnReconnect` isn't in ADR 0006's option list — the ADR's **Why** section justifies `retry: false`, `networkMode: 'always'`, and `refetchOnWindowFocus: false` only. The reasoning in the comment here is sound and the changeset describes the behaviour; the ADR just never picked it up.

Since the ADR lands in this same branch, amending it is a one-paragraph fix rather than a follow-up.

packages/hooks/src/internal/useQueryKeyBase.ts
```diff
@@ -0,0 +1,44 @@
...
+    () => serializeAdditionalHeaders(additionalHeaders),
+    [additionalHeaders],
+  );
+
+  return [
+    context?.appKey,
+    context?.apiHost,
+    context?.installationId,
+    additionalHeadersKey,
+    serializeVersionFilters(),
```
bmanquen, line 49: **issue (non-blocking):** The version-filter key segment appears in neither ADR 0006 nor the changeset. Worth a line in the ADR alongside `refetchOnReconnect`.

**question:** `serializeVersionFilters()` reads `YouVersionPlatformConfiguration` at render rather than context. `YouVersionProvider` assigns those statics during render so the first child fetch sees them, but nothing subscribes — a host mutating `permittedVersionIds` on the config directly changes the key with no re-render to observe it, so the new key isn't picked up until something else re-renders. Is that path considered supported, or is provider-props-only the contract?

bmanquen (commented):
Non-blocking follow-up to the two blockers above: repo-standard breaches, duplication, and three places the code drifted from ADR 0006 (which this branch adds, so the ADR can just be amended). Posted by Claude Code, an AI agent, on behalf of Brenden.

packages/hooks/src/useApiData.ts
```diff
@@ -1,95 +1,116 @@
...
+  //
+  // If `enabled` is false, `loading` is `false`.
+  // TanStack Query keeps `isPending` true.
+  return {
+    data: enabled ? (query.data ?? null) : null,
+    loading:
+      enabled &&
+      (query.isPending ||
+        (query.isPlaceholderData && query.isFetching) ||
+        (query.isError && query.isFetching)),
```
bmanquen, line 118: **issue (non-blocking):** This third `loading` clause contradicts ADR 0006 in two places. The **Consequences** section defines `loading` as "the first load of a key (`isPending`)" plus "the placeholder window after a key change (`isPlaceholderData && isFetching`)" — two clauses, not three. And accepted delta (1) states "`refetch` with cached data no longer flips `loading` — data updates in place," which this breaks for the errored-with-cached-data case that delta (2) creates.

The behaviour looks like the better one (a `refetch` from an error state should show progress). It's the ADR that needs updating, not the code.

packages/hooks/src/internal/useUserScope.ts
```diff
@@ -0,0 +1,53 @@
...
+ * each path that changes the token also changes the auth context, and the
+ * context change re-renders this hook.
+ *
+ * This hook reads `YouVersionAuthContext` directly. It does not call
+ * `useYouVersionAuthContext()`, because that call throws when auth is off,
+ * and data hooks must work without auth. No auth context means that the host
+ * owns identity, so one scope covers any token the host sets. The React
+ * Native Expo SDK also puts `userInfo` in this context.
+ */
+export function useUserScope(): string | null {
```
bmanquen, line 42: **nitpick (non-blocking):** None of the six modules in `src/internal/` carry `/** @internal */`, which `CONTRIBUTING.md` asks for on internal APIs. They're absent from `src/index.ts` so nothing leaks into the `.d.ts` today — raising it only because this branch is where the `internal/` directory grew, and the tag is what keeps a future accidental re-export from becoming public surface.

packages/hooks/src/useApiData.ts
```diff
@@ -1,95 +1,116 @@
...
+ * Then TanStack Query fetches a new copy.
+ */
+export function useApiData<TData>(
+  queryKey: readonly unknown[],
+  fetchFn: () => Promise<TData>,
   options: UseApiDataOptions = {},
-): UseApiDataResult<T> {
-  const { enabled = true } = options;
+): UseApiDataResult<TData> {
+  const { enabled = true, keepPreviousData: holdPreviousData = true } = options;
```
bmanquen: **nitpick (non-blocking):** `keepPreviousData: holdPreviousData` gives one concept two names inside a 116-line file, and the rename only buys shadow-avoidance against the imported TanStack `keepPreviousData` symbol. Aliasing the import instead would keep the option and the local reading the same.

packages/hooks/src/context/YouVersionProvider.tsx
cameronapak: Hey, is this asking for the consumer to peer dep in or use tanstack query for themselves?

Dustin-Kelley: **note:** Fixed in `8607554`. To Cam's question: neither — TanStack Query stays a direct, exact-pinned dependency of the hooks package. Consumers install nothing and remain free to use (or not use) TanStack Query themselves.

The fix is that `YouVersionProvider` no longer renders TanStack's `QueryClientProvider` at all. The private client travels on a package-private context and reaches `useQuery`/`useQueries` as the explicit client argument, so TanStack's own context is never written or read in either direction — no shadowing, regardless of how the host's `@tanstack/react-query` import dedupes. The `client.mount()`/`unmount()` lifecycle that `QueryClientProvider` used to run (which `refetchOnReconnect` depends on) is preserved by the internal provider. Test seams that render the context raw mount `TestQueryClientProvider` from the published `@youversion/platform-react-hooks/test-utils` subpath.

---
Posted by Claude Code, an AI agent, on behalf of Dustin.

packages/hooks/src/internal/useUserScope.ts
Dustin-Kelley: **note:** Fixed in `2f795eb`. The no-context branch now runs the same token check as the branch below it: an access token set on `YouVersionPlatformConfiguration` while no identity is known returns `null` (no fetch), and `'anon'` is reserved for the genuinely signed-out state. The docstring now says token-only identity is unsupported outright — hosts that set tokens directly must render the auth provider (`includeAuth`) or supply `userInfo` through the context (the RN Expo path, which stays safe as you noted).

---
Posted by Claude Code, an AI agent, on behalf of Dustin.

packages/ui/src/components/bible-reader-highlights-machine.ts
Dustin-Kelley: **note:** Fixed in `d6db621` — this path now also checks, so the spec line about defusing in-flight writes holds. `settleWrite` computes whether the op still owns any of its verses (read from the context before the assign releases the op's claims) and the 401/403 path stashes the pending highlight and raises `PERMISSION_LOST` only when it does. A sign-out, account change, or scope change clears `writeIntent`, so a write from account A that settles after an A-to-B swap owns nothing and opens no dialog over account B. The permission-cache invalidation and refetch on that path stay ungated — they are conservative and target the current account's state. New regression test: a disowned write that 401s after an account swap stashes nothing and opens no dialog. The host A-to-B swap deferral for in-flight reads is unchanged.

---
Posted by Claude Code, an AI agent, on behalf of Dustin.

packages/hooks/src/internal/queryClientDefaults.ts
Dustin-Kelley: **note:** Fixed in `d70d1c4`. ADR 0006 now names both rules: the Why section has a `refetchOnReconnect` bullet naming the predicate (`query.state.status === 'error'`, errored reads only, with the recovery rationale), and Consequences defines `loading` as three windows including `isError && isFetching`, with accepted delta (1) qualified so it no longer contradicts the third clause. The `queryClientDefaults.ts` pointer to the ADR is accurate again.

---
Posted by Claude Code, an AI agent, on behalf of Dustin.

packages/hooks/src/internal/useQueryKeyBase.test.tsx
Dustin-Kelley: **note:** Fixed in `95b3cbe`. The file is now one top-level `describe` with flat tests and no `afterEach`; the two tests that mutate the version-filter statics run their bodies through a `withVersionFilterCleanup` try/finally wrapper that restores the statics however the body exits.

---
Posted by Claude Code, an AI agent, on behalf of Dustin.

packages/hooks/src/useApiData.test.tsx
Dustin-Kelley: **note:** Fixed in `95b3cbe`. Same treatment as the other file: one flat top-level `describe`, both `afterEach` blocks gone. Tests that touch `onlineManager` or `focusManager` run through `withOnlineCleanup`/`withFocusCleanup` try/finally wrappers that hand the manager back however the body exits, and the `refocus` helper moved to module scope.

---
Posted by Claude Code, an AI agent, on behalf of Dustin.

packages/ui/src/test/hook-overrides.tsx
Dustin-Kelley: **note:** Fixed in `8607554`. The mirror is gone: the ui test provider is now a re-export of `TestQueryClientProvider` from `@youversion/platform-react-hooks/test-utils`, which builds its client from the hooks package's own `queryClientDefaultOptions` (also re-exported there). One source of truth; the "change both copies" docstring is deleted.

---
Posted by Claude Code, an AI agent, on behalf of Dustin.

packages/ui/src/test/setup.ts
Dustin-Kelley: **note:** Fixed in `ca01f2c`. The duplicated graft block is now `graftTestWebStorage()` imported from `@youversion/platform-react-hooks/test-utils`; only ui's `scrollIntoView` stub (and its cleanup `afterEach` from testing-library) remains local.

---
Posted by Claude Code, an AI agent, on behalf of Dustin.

packages/hooks/src/useHighlights.ts
Dustin-Kelley: **note:** Fixed in `79597a9`. `null` now goes into the key as-is — the query is disabled for exactly that render, so the key is never fetched, and the comment explains why a sentinel string would only suggest the entry is real.

---
Posted by Claude Code, an AI agent, on behalf of Dustin.

packages/hooks/src/internal/queryClientDefaults.ts
Dustin-Kelley: **note:** Fixed in `d70d1c4`. ADR 0006's Why section now has a `refetchOnReconnect` bullet naming the predicate (`query.state.status === 'error'`), why `networkMode: 'always'` would otherwise infer `false`, and why reconnect is an errored read's only automatic way back with retries and focus refetch off.

---
Posted by Claude Code, an AI agent, on behalf of Dustin.

packages/hooks/src/internal/useQueryKeyBase.ts
Dustin-Kelley: **note:** Fixed in `d70d1c4` — the ADR's Decision section now names the serialized version filter as a key-base segment, with the tightened-filter rationale.

On the question: provider-props-only is the contract. `useQueryKeyBase`'s docstring now states it — the provider syncs its props onto the config during render, before children read the value, so every child render sees the current filter; a direct write to the config statics from app code triggers no render and is unsupported (it would be picked up on the next render for any other reason, not at the moment of the write).

---
Posted by Claude Code, an AI agent, on behalf of Dustin.

packages/hooks/src/useApiData.ts
Dustin-Kelley: **note:** Fixed in `d70d1c4`. ADR Consequences now defines `loading` as three windows — first load (`isPending`), the placeholder window (`isPlaceholderData && isFetching`), and an errored read's in-flight retry (`isError && isFetching`) — with the rationale that a recovering hook must not render the error state as settled. Accepted delta (1) is qualified to settled, non-errored data so it no longer contradicts the third clause.

---
Posted by Claude Code, an AI agent, on behalf of Dustin.

packages/hooks/src/internal/useUserScope.ts
Dustin-Kelley: **note:** Fixed in `9350707`. All six `src/internal/` modules now carry `@internal` on their exported symbols' docstrings, matching `QueryClientContext`.

---
Posted by Claude Code, an AI agent, on behalf of Dustin.

packages/hooks/src/useApiData.ts
Dustin-Kelley: **note:** Fixed in `3b08bd0`. The destructured option keeps the name callers pass — `keepPreviousData` — and the TanStack placeholder import carries the alias instead.

---
Posted by Claude Code, an AI agent, on behalf of Dustin.

cameronapak (approved):
Greppy seems happy. I'm happy

PR #382, opened 2026-09-03
perf: cut published package and partner-bundle bytes (YPE-5528)

cameronapak:
## Summary

- Make package-root UI imports tree-shakeable without changing the public API or adding subpaths.
- Keep core external so UI and hooks share one runtime copy.
- Split Provider, component, and reader styles so consumers only bundle what they use.
- Keep existing eager i18next localization behavior.

## Bundle impact

Minified Brotli, esbuild without splitting, compared with `main` (`90374fe`):

| Import | `main` | This PR | Change |
| --- | ---: | ---: | ---: |
| `YouVersionProvider` | 152.3 kB | 48.9 kB | −68% |
| `BibleReader` | 140.4 kB | 142.1 kB | +1.3% |
| `BibleCard` | 144.6 kB | 105.7 kB | −27% |
| `Separator` | 140.4 kB | 16.8 kB | −88% |
| Full barrel | 161.4 kB | 158.4 kB | −1.9% |

## Verification

Forced build, typecheck, 523 unit tests, 569 browser/unit tests, tree-shaking checks, size limits, and the Vite consumer build pass.

YPE-5528. Parent: YPE-1193.

<h3>Greptile Summary</h3>

This PR restructures the UI package into tree-shakeable component entries while preserving package-root exports and sharing the core runtime with hooks.
- Splits provider, component, and reader styles into independently injected resources while retaining a complete public stylesheet.
- Adds build-time checks for public entry/export parity, stylesheet integrity, tree-shaking, and consumer bundle budgets.
- Updates Storybook, release/CDN publishing, documentation, and package metadata for the new output layout.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/ui/tsup.config.ts | Defines split public component entries, externalizes the shared core runtime, and embeds the three generated style resources. |
| packages/ui/package.json | Reorders and validates clean UI builds, exposes the composed stylesheet, and aligns publishing with the core version stamp. |
| packages/ui/src/components/YouVersionProvider.tsx | Removes synchronization for the formerly inlined core copy and limits provider styling to the lightweight chrome sheet. |
| packages/ui/src/i18n/index.ts | Uses eagerly bundled locale resources, eliminating the asynchronous locale-chunk loading paths behind the two previous findings. |
| packages/ui/scripts/check-public-entries.mjs | Guards against drift between public component exports and explicit tsup entries. |
| packages/ui/scripts/verify-styles.js | Validates all generated, embedded, and publicly exported style artifacts. |
| scripts/check-tree-shaking.mjs | Adds consumer-shaped verification that narrow UI imports exclude unrelated component sentinels while retaining required styles. |
| .github/workflows/release.yml | Resolves the CDN upload artifact from the package’s public stylesheet export. |

<h3>Flowchart</h3>

```mermaid
flowchart LR
  Root[Package-root named import] --> Entry[Per-component build entry]
  Entry --> Core[Shared external core runtime]
  Entry --> Chrome[Provider chrome styles]
  Entry --> Components[Component utility styles]
  Entry --> Reader[Reader typography styles]
  Components --> Public[Complete public styles.css]
  Reader --> Public
  Public --> CDN[Versioned CDN bible.css]
```

<sub>Reviews (13): Last reviewed commit: ["Merge branch &#39;main&#39; into cp/test-small-p..."](https://github.com/youversion/platform-sdk-react/commit/42d2d58c9993e37e0ce8a05190ad03471860f28d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=61719122)</sub>

**Context used:**

- Knowledge Base — [UI provider and styling system](https://app.greptile.com/youversion/-/custom-context/knowledge-base/youversion/platform-sdk-react/-/docs/ui-provider-styling.md)
- Knowledge Base — [Package delivery workflows](https://app.greptile.com/youversion/-/custom-context/knowledge-base/youversion/platform-sdk-react/-/docs/delivery-workflows.md)


packages/ui/src/lib/yv-styles-components.tsx
cameronapak, line 1: nitpick: I want some information on the JSDoc regarding how React 19 has de-dupe support out of the box when the href is the same

packages/ui/src/lib/yv-styles-reader.tsx
cameronapak, line 1: nitpick: I want some information on the JSDoc regarding how React 19 has de-dupe support out of the box when the href is the same

packages/ui/src/lib/yv-styles-components.tsx
cameronapak: resolved: Added JSDoc in [0628604](https://github.com/youversion/platform-sdk-react/commit/062860480353144515d5daa7b552b11f341dcccf) explaining that React 19 hoists the sheet into `<head>` and de-duplicates repeated renders by matching `href`.

Written by Cursor, an AI agent sent on behalf of Cam.

packages/ui/src/lib/yv-styles-reader.tsx
cameronapak: resolved: Added the same React 19 contract in [0628604](https://github.com/youversion/platform-sdk-react/commit/062860480353144515d5daa7b552b11f341dcccf): repeated reader style renders share one hoisted style element because they use the same `href`.

Written by Cursor, an AI agent sent on behalf of Cam.

Dustin-Kelley (commented):
_Posted by Claude Code, an AI agent sent on behalf of Dustin._

The refactor holds up well overall: chrome/component/reader sheets have token parity with `theme.css`, the hooks provider covers the removed `YouVersionPlatformConfiguration` mirroring, every Bible HTML path goes through `BibleTextView` so the reader sheet is always injected, and ESM/CJS export parity plus core re-export coverage passed on a real build. Three inline comments below, plus one on a file outside the diff:

**nitpick (non-blocking):** `scripts/check-sdk-version-stamp.mjs` — the header comment still says UI bundles core via `noExternal`, and the `ui` mode would now fail closed with "no SDK version code found" since the only caller switched to `core`. Worth updating the comment and either removing `ui` mode or pointing it at the core dist.

packages/ui/src/styles/global.css
```diff
@@ -49,10 +49,14 @@
 layer(yv-sdk-fonts);
 @import '@youversion/platform-core/browser/styles/theme.css' layer(yv-sdk-styles);
 @import 'tailwindcss/theme.css' layer(yv-sdk-theme) prefix(yv);
-@import 'tailwindcss/utilities.css' layer(yv-sdk-utilities) prefix(yv);
-@import '@youversion/platform-core/browser/styles/bible-reader.css' layer(yv-sdk-bible-reader);
+@import 'tailwindcss/utilities.css' layer(yv-sdk-utilities) prefix(yv) source(none);
 @import 'tw-animate-css';
 
+@source '../components';
+@source not '../**/*.stories.tsx';
```
Dustin-Kelley: **issue (blocking):** `@source not '../**/*.stories.tsx'` strips story-only utilities from `dist/tailwind.css`, which is the only sheet Storybook embeds (`__YV_COMPONENT_STYLES__`).

I built the branch and diffed the class names used in `*.stories.tsx` against the emitted sheet — 10 of 37 are missing: `yv:h-screen`, `yv:h-svh`, `yv:max-h-svh`, `yv:items-end`, `yv:p-8`, `yv:p-12`, `yv:max-w-md`, `yv:max-w-lg`, `yv:w-4`, `yv:grid-rows-[auto_1fr]`. Every `bible-reader.stories.tsx` wrapper is `<div data-yv-sdk className="yv:h-screen">` with a `yv:h-full` root beneath it, so the reader collapses to content height; the chapter/version picker stories lose their `yv:items-end` bottom anchoring; `verse.stories.tsx:507` loses its `yv:h-svh yv:max-h-svh` scroll container.

Suggest either scanning stories into a Storybook-only sheet in `viteFinal`, or moving the story wrappers to inline styles so the published sheet can stay narrow.

_— Claude Code, on behalf of Dustin_

packages/ui/package.json
```diff
@@ -33,17 +33,19 @@
...
     "verify:styles": "node scripts/verify-styles.js",
     "generate:i18n": "node scripts/generate-i18n-resources.mjs",
-    "build": "pnpm generate:i18n && pnpm build:css && pnpm build:js && pnpm verify:styles && pnpm build:types",
-    "dev": "concurrently \"node scripts/generate-i18n-resources.mjs --watch\" \"tailwindcss -i src/styles/global.css -o dist/tailwind.css --watch\" \"NODE_ENV=development tsup --watch\"",
+    "check:entries": "node scripts/check-public-entries.mjs",
+    "check:exports": "node scripts/check-public-exports.mjs",
+    "build": "rm -rf dist && pnpm generate:i18n && pnpm check:entries && pnpm build:css && pnpm build:js && pnpm check:exports && pnpm verify:styles && pnpm build:types",
+    "dev": "concurrently \"node scripts/generate-i18n-resources.mjs --watch\" \"tailwindcss -i src/styles/chrome.css -o dist/chrome.css --watch\" \"tailwindcss -i src/styles/global.css -o dist/tailwind.css --watch\" \"tailwindcss -i src/styles/reader.css -o dist/bible-reader.css --watch\" \"NODE_ENV=development tsup --watch\"",
```
Dustin-Kelley: **issue (non-blocking):** `dev` (and `storybook` on L48) never produce `dist/styles.css`, but that's now the target of the `./styles.css` export and the README's recommended import.

Only `build:css` runs `scripts/write-public-styles.js`; the watch scripts emit `chrome.css`, `tailwind.css`, and `bible-reader.css` and stop. From a clean checkout (or after `build`'s `rm -rf dist`), `pnpm --filter @youversion/platform-react-ui dev` leaves `import '@youversion/platform-react-ui/styles.css'` unresolvable until a full `build` runs. Before this PR the export pointed at `dist/tailwind.css`, which the watcher did write.

Suggest running `write-public-styles.js` once before the watchers start, or chaining it after the `tailwind.css` watcher via a small concat step.

_— Claude Code, on behalf of Dustin_

packages/ui/src/components/YouVersionProvider.tsx
```diff
@@ -1,8 +1,9 @@
+'use client';
```
Dustin-Kelley, line 1: **question (non-blocking):** Are the new `'use client'` directives (here and in AuthButton, BibleCard, BibleVersionPicker, VerseActionPopover, VerseOfTheDay) intended to take effect in this PR? They don't reach `dist/` — I grepped ESM, CJS, and `components/*.js` after building and found zero occurrences; esbuild drops them under `splitting: true`.

`main` didn't ship them either, so this isn't a regression, but the diff reads as though it adds RSC boundaries when it doesn't — a Next.js App Router consumer importing `YouVersionProvider` from a server component still hits the hooks error. If the intent is real, tsup needs a `banner` on entry outputs or a directive-preserving plugin; if not, dropping the directives avoids implying a guarantee the build doesn't deliver.

_— Claude Code, on behalf of Dustin_

packages/ui/src/styles/global.css
cameronapak: resolved: Added a Storybook-only Tailwind sheet in [7b912e5](https://github.com/youversion/platform-sdk-react/commit/7b912e5f1fa28dbbf26ed6303d091bc05f60c0ac), so story wrapper utilities are available without widening the published component sheet. All ten missing utilities are generated, and the BibleReader and picker layouts were verified in the rendered Storybook.

packages/ui/src/components/YouVersionProvider.tsx
cameronapak: answer: Yes, the client boundaries are intended. In [7b912e5](https://github.com/youversion/platform-sdk-react/commit/7b912e5f1fa28dbbf26ed6303d091bc05f60c0ac), the build restores `use client` after tsup’s final bundling pass for every ESM and CommonJS output, and the tree-shaking check guards that contract.

cameronapak:
resolved: Updated the stale bundling explanation and removed the unused `ui` mode in [7b912e5](https://github.com/youversion/platform-sdk-react/commit/7b912e5f1fa28dbbf26ed6303d091bc05f60c0ac). UI now keeps core external, so both package prepublish paths verify the single core runtime directly.

packages/ui/package.json
cameronapak: resolved: The dev and Storybook commands now build the CSS outputs before starting in [7b912e5](https://github.com/youversion/platform-sdk-react/commit/7b912e5f1fa28dbbf26ed6303d091bc05f60c0ac), and the public stylesheet writer watches the component and reader sheets so `dist/styles.css` stays resolvable and current.

packages/ui/.storybook/styles.css
```diff
@@ -0,0 +1,5 @@
+/* Story wrappers use utilities that do not belong in the published component sheet. */
+@import 'tailwindcss/theme.css' layer(yv-sdk-theme) prefix(yv);
```
Dustin-Kelley: **issue (blocking):** The Storybook sheet re-emits component utilities with stock Tailwind token values, and because `main.ts` appends it after `dist/tailwind.css` in the same `yv-sdk-utilities` layer, those copies win.

This file imports plain `tailwindcss/theme.css` and `@source`s all of `../src`, so every `yv:` class used by components is generated a second time — but without `global.css`'s `@theme inline` overrides. I built both sheets from `16a7d2a`:

- `dist/tailwind.css`: `.yv\:font-serif{font-family:Untitled Serif,"Source Serif 4",serif}`, `.yv\:font-sans{font-family:Inter,sans-serif}`, `.yv\:rounded-md{border-radius:calc(var(--yv-radius) - 2px)}`
- `generated-styles.css`: `.yv\:font-serif{font-family:var(--yv-font-serif)}` with `--yv-font-serif:ui-serif,Georgia,…` and `--yv-radius-md:.375rem` declared in `@layer yv-sdk-theme`

Net effect in Storybook: reader and verse stories render in Georgia instead of Untitled Serif / Source Serif 4, Inter falls back to `system-ui`, and card radii lose the `--yv-radius` token — so Storybook no longer shows the shipped typography (Swift is the source of truth for reader type, per AGENTS.md).

Suggest narrowing to `@source '../src/**/*.stories.tsx'` and importing `../src/styles/global.css` (or duplicating its `@theme inline` block) so any overlapping utilities compile to identical declarations.

_— Claude Code, on behalf of Dustin_

packages/ui/tsup.config.ts
```diff
@@ -1,38 +1,83 @@
...
     '.css': 'text',
   },
   format: ['esm', 'cjs'],
+  // Rollup strips module-level directives. Every public UI entry uses React
+  // client APIs, so restore the package boundary after the final bundling pass.
+  plugins: [
+    {
+      name: 'restore-use-client-directives',
+      buildEnd({ writtenFiles }) {
+        for (const { name } of writtenFiles) {
```
Dustin-Kelley, line 45: **issue (non-blocking):** Prepending `'use client'` to every output includes `dist/index.js`, which turns the barrel's non-component re-exports into client references under RSC.

`index.ts` re-exports core values like `ApiClient`, `BibleClient`, `YouVersionPlatformConfiguration`, `transformBibleHtml`, `SDK_VERSION`. In a `'use client'` module every export is a client reference on the server, so a Next.js App Router consumer doing `import { YouVersionPlatformConfiguration, BibleClient } from '@youversion/platform-react-ui'` in a server component or route handler now gets proxies: `YouVersionPlatformConfiguration.appKey = …` writes to the proxy and `new BibleClient()` throws "Attempted to call … from the server but … is on the client". Before `7b912e5` no directive was emitted, so those imports worked server-side. `assertUiClientDirectives` in `scripts/check-tree-shaking.mjs` also now fails the build if `index.js` loses the directive, locking this in.

Suggest scoping the plugin to the `components/*` entries and their chunks and leaving `index.js` undirected — or, if the barrel is meant to be client-only, say so in the README and point server code at `@youversion/platform-core`.

_— Claude Code, on behalf of Dustin_

Dustin-Kelley (commented):
_Posted by Claude Code, an AI agent sent on behalf of Dustin._

Re-reviewed the follow-up commits (`7b912e5` … `8e7c7e7`). All three earlier threads are resolved as described: the Storybook sheet exists, `dev`/`storybook` run `build:css` up front plus a `--watch` mode of `write-public-styles.js`, the tsup plugin really does restore `'use client'` (`writtenFiles.name` is cwd-relative, so the read/write works), and `check-sdk-version-stamp.mjs` is cleaned up. Two new inline comments from the fixes, plus two small notes:

**chore (non-blocking):** `AGENTS.md` drops the "identify as Cursor on behalf of Cam" preference line — unrelated to YPE-5528; intentional?

**nitpick (non-blocking):** `.changeset/silly-ducks-push.md` is an empty changeset alongside the existing `ype-5528-ui-split-dist.md`; harmless but redundant.

packages/ui/.storybook/styles.css
cameronapak: resolved: Fixed in [7d0b3d3](https://github.com/youversion/platform-sdk-react/commit/7d0b3d39bce8f7f4d1fec9dd78b0d09a3b9d28a4). Storybook now builds one component sheet from the production `global.css` plus story-only sources, so overlapping utilities retain the shipped font and radius tokens instead of overriding them. All ten story-only utilities are present, and the rendered reader and picker layouts were verified.

packages/ui/tsup.config.ts
cameronapak: resolved: Fixed in [7d0b3d3](https://github.com/youversion/platform-sdk-react/commit/7d0b3d39bce8f7f4d1fec9dd78b0d09a3b9d28a4). The build now restores `use client` only to component entries and their chunks; `index.js` and `index.cjs` stay unmarked so package-root core re-exports remain usable on the server. The tree-shaking gate asserts both sides of that boundary.

cameronapak:
answer: Both files came from `main` via [#411](https://github.com/youversion/platform-sdk-react/pull/411), not from this PR. That PR intentionally removed the review identity instruction and added the empty changeset; neither file appears in PR #382’s current diff.

Dustin-Kelley (approved):
🔥

PR #384, opened 2026-09-08
perf(core): split clients into tree-shakable modules and cut published bytes

cameronapak:
Stack 1/3 — base of #382.

## What
- Core clients split into `bible-*` / `languages-*` modules so partners importing one symbol stop pulling the whole client graph
- Zod-mini schemas, minified browser CSS, shared filtered paged-list helper (`fetchFilteredCollection`)
- Version-stamp check updated for the new dist layout

## Why
Heart goal of the branch: smaller import size for partners, backwards compatible.

## Verification
- `pnpm build --force`, `pnpm typecheck`, `pnpm test` (432 core tests), `pnpm check:tree-shaking` all pass on this branch alone — with **main's** hooks and ui, which is the back-compat proof
- No public API change; lockfile unchanged

Merge first of the stack (then #385, then #382).

— Cursor, an AI agent sent on behalf of Cam

<h3>Greptile Summary</h3>

The PR splits core clients into narrower tree-shakable modules and reduces published output while preserving the existing public surface.
- Moves client input contracts into schema modules and adopts `zod/mini`.
- Adds generated, minified browser CSS output while retaining public theme tokens.
- Updates build, development, package-export, and version-stamp handling for the new distribution layout.
- Adds a patch changeset documenting the validation-error identity change.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/core/src/styles/theme.css | Restores the public expressive palette, chart, and sidebar tokens identified by the previous review. |
| packages/core/src/languages-list.ts | Uses and re-exports the schema-derived language-options contract, resolving the previous schema-first violation. |
| packages/core/src/schemas/language.ts | Defines runtime validation and the inferred public type for language-list options. |
| packages/core/package.json | Redirects browser-style exports to generated distribution assets and configures the revised build pipeline. |
| .changeset/ype-5528-core-split.md | Records the patch release and explicitly documents the accepted zod Mini validation-error behavior. |
| packages/core/src/bible-passage.ts | Extracts passage retrieval and HTML transformation behavior from the prior Bible client implementation. |

<sub>Reviews (8): Last reviewed commit: ["fix(core): drop ZodError wrap and docume..."](https://github.com/youversion/platform-sdk-react/commit/4a4d74eebc454cfac3b7a3d10b4eebacbbb1516b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=61718724)</sub>

<details><summary><h4>Context used (4)</h4></summary>

- Knowledge Base — [Core platform SDK](https://app.greptile.com/youversion/-/custom-context/knowledge-base/youversion/platform-sdk-react/-/docs/core-platform.md)
- Knowledge Base — [Bible HTML transformation and browser styles](https://app.greptile.com/youversion/-/custom-context/knowledge-base/youversion/platform-sdk-react/-/docs/core-html-styling.md)
- Knowledge Base — [Monorepo build, validation, and release](https://app.greptile.com/youversion/-/custom-context/knowledge-base/youversion/platform-sdk-react/-/docs/monorepo-build-release.md)
- Knowledge Base — [Package delivery workflows](https://app.greptile.com/youversion/-/custom-context/knowledge-base/youversion/platform-sdk-react/-/docs/delivery-workflows.md)
</details>


packages/core/src/styles/theme.css
cameronapak, line 1: issue: lots of CSS removed. Let's keep the CSS tokens/variables please.

packages/core/package.json
```diff
@@ -52,7 +81,8 @@
     "dev": "tsup --watch",
     "build:js": "tsup",
     "build:types": "tsc -p tsconfig.build.json",
-    "build": "pnpm build:js && pnpm build:types",
+    "build:css": "node scripts/minify-browser-css.mjs",
+    "build": "rm -rf dist && pnpm build:js && pnpm build:css && pnpm build:types",
     "prepublishOnly": "YVP_PUBLISH_BUILD=true pnpm build && node ../../scripts/check-sdk-version-stamp.mjs core",
```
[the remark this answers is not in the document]
cameronapak: Changeset added

packages/core/src/styles/theme.css
cameronapak: Fixed

scripts/check-sdk-version-stamp.mjs
```diff
@@ -21,7 +21,6 @@ const repoRoot = resolve(dirname(fileURLToPath(import.meta.url)), '..');
 
 const PACKAGE_DIRS = {
   core: 'packages/core',
```
abharms, line 24: issue (blocking): Removing the `ui` entry makes the UI package impossible to publish. Its `prepublishOnly` script still runs `node ../../scripts/check-sdk-version-stamp.mjs ui`, so the publish lifecycle deterministically exits with `unknown package "ui"`. Since UI still bundles core through `noExternal`, please retain the UI artifact check until that bundling changes.

— Cursor, an AI agent sent on behalf of Cam.

packages/core/package.json
```diff
@@ -52,7 +81,8 @@
     "dev": "tsup --watch",
```
abharms: issue: Now that the public CSS exports resolve from generated `dist/styles`, this watch task also needs to rebuild/watch CSS. Turbo performs the dependency build once, but subsequent edits to `packages/core/src/styles/*.css` never update `dist/styles`, so the UI Tailwind watcher continues serving stale core styles during `pnpm dev`.

— Cursor, an AI agent sent on behalf of Cam.

packages/core/src/schemas/language.ts
```diff
@@ -1,38 +1,92 @@
...
+  .check(
+    z.refine(
+      (data) => {
+        if (data?.page_size === '*') {
+          return data.fields && data.fields.length >= 1 && data.fields.length <= 3;
+        }
+        return true;
+      },
+      {
+        error: 'page_size="*" required 1-3 fields to be specified',
```
abharms: issue (blocking): This moves the `page_size="*"` guard from the client’s explicit `Error` into Zod validation. Existing callers now receive a `$ZodError` whose `.message` is JSON-shaped instead of the prior human-readable `page_size="*" requires 1-3 fields to be specified`; the text also changes `requires` to `required`, contradicting the changeset’s claim that custom messages are unchanged. Please preserve the prior observable error contract or classify and document the break accordingly.

— Cursor, an AI agent sent on behalf of Cam.

packages/core/src/parse-public.ts
```diff
@@ -0,0 +1,39 @@
...
+
+function throwAsClassicZodError(cause: unknown): never {
+  if (cause instanceof z.core.$ZodError) {
+    throw new ClassicZodError(cause.issues);
+  }
+  throw cause;
+}
+
+/** Parse at a public client boundary, rethrowing Mini failures as classic ZodError. */
+export function parsePublic<Input, Output>(
```
cameronapak: **[nitpick](https://conventionalcomments.org/):** Document that public schema failures guarantee `instanceof ZodError` plus `.issues` / `.message`, not classic helpers like `format()` / `flatten()`.

Greptile’s follow-up matches: the wrap restores identity, not the full classic method surface.

<details>
<summary>For Agents</summary>

Standards important (non-blocking). Optional: proxy helpers or narrow changeset wording. Not a Cam-selected must-fix.

</details>

By Code Reviewer bot, sent on behalf of Cam.

cameronapak (commented):
## Review

No Jira. Spec is the Cam-selected review-comment fixes (Conductor re-review).

## Summary

- **Standards must-fix:** 0
- **Spec must-fix:** 0
- **Worst:** none blocking (ZodError wrap covers `instanceof`; classic helpers like `format()`/`flatten()` are still absent)

<details>
<summary>For Agents</summary>

CI: pass (Lint Commit Messages may still be settling)
Greptile: pass / partnered — old ZodError P1 addressed for `instanceof`; residual method-surface caveat noted below
Event: COMMENT (Cam-authored — no self-APPROVE)
HEAD: `4703196`
Cam fixes verified:
1. `ui` in `PACKAGE_DIRS` — pass
2. `page_size="*"` exact `Error` text — pass
3. `dev.mjs` CSS watch → `dist/styles` — pass
4. changeset patch + `parsePublic` Mini wrap — pass
Resolve outdated Austin/Cursor threads for 1–3 when posting.

</details>

By Code Reviewer bot, sent on behalf of Cam.

.changeset/ype-5528-core-split.md
```diff
@@ -0,0 +1,7 @@
...
+
+Behavior note: input-validation failures throw zod Mini's `$ZodError` (from `zod/mini`) instead of classic `ZodError`. `instanceof ZodError` checks against classic `zod` no longer match. Callers that catch validation errors should match on `error instanceof $ZodError` from `zod/mini` (or `z.core.$ZodError`). `page_size="*"` without 1-3 fields still throws the prior `Error` (`page_size="*" requires 1-3 fields to be specified`).
```
cameronapak, line 7: **[nitpick](https://conventionalcomments.org/):** Lead the catch guidance with `z.core.$ZodError` after `import * as z from 'zod/mini'`. Mini does not export a top-level `$ZodError`.

<details>
<summary>For Agents</summary>

Standards important (non-blocking). Primary `instanceof $ZodError` from `zod/mini` misleads; parenthetical is the real path.

</details>

By Code Reviewer bot, sent on behalf of Cam.

cameronapak (commented):
## Review

No Jira. Spec is the Cam override (drop `parsePublic` wrap; document Mini `$ZodError`).

## Summary

- **Standards must-fix:** 0
- **Spec must-fix:** 0
- **Worst:** none blocking (changeset catch snippet should prefer `z.core.$ZodError`)

<details>
<summary>For Agents</summary>

CI: pass
Greptile: pass
Event: COMMENT (Cam-authored — no self-APPROVE)
HEAD: `9ff6bbd`
Verified: parsePublic gone; `.parse()`; changeset patch + Mini note; ui stamp; page_size Exact Error; CSS watch.

</details>

By Code Reviewer bot, sent on behalf of Cam.

packages/core/src/parse-public.ts
cameronapak: Closing as outdated.

`parse-public.ts` and the classic `ZodError` wrap were removed per Cam’s override. Schema failures throw Mini `$ZodError` only. The `format()` / `flatten()` note no longer applies to a wrap that does not exist.

No further code change for this thread.

.changeset/ype-5528-core-split.md
cameronapak: Acknowledged; non-blocking nit.

Cam is taking this PR to Austin for re-review without another docs-only pass. Catch guidance in `.changeset/ype-5528-core-split.md` can lead with `z.core.$ZodError` after `import * as z from 'zod/mini'` in a follow-up if Austin wants it.

Resolving so the thread list is clean for that re-review.

abharms (approved):
💯

docs/adr/0002-sdk-version-dev-suffix-on-build.md
```diff
@@ -51,10 +51,10 @@ Mechanism (`packages/core`):
...
   channel cannot be confirmed aborts the publish.
 
-`@youversion/platform-react-ui` bundles core (`noExternal`), so it inherits the
-stamp from the core dist it bundles; its `prepublishOnly` runs the same guard.
-`@youversion/platform-react-hooks` imports core at runtime and bakes no version,
-so it needs no change.
+`@youversion/platform-react-ui` and `@youversion/platform-react-hooks` import
+core at runtime and bake no version. The stamp lives in published core. UI
+`prepublishOnly` rebuilds core with `YVP_PUBLISH_BUILD=true` and runs the same
+guard on core.
```
Dustin-Kelley, line 57: **issue:** this paragraph is factually wrong on two counts, and contradicts a comment added in this same PR.

1. *"`@youversion/platform-react-ui` … import[s] core at runtime and bake[s] no version."* — `packages/ui/tsup.config.ts:22` sets `noExternal: ['@youversion/platform-core']`, so ui inlines core and **does** bake the constant. A clean dev build of ui puts `isPublishBuild = false` directly into `dist/index.js` and `dist/index.cjs`; running the guard against it reports:

   ```
   check-sdk-version-stamp: packages/ui would publish an unstamped (-dev) X-YVP-Sdk header.
   Affected files: index.cjs, index.js
   ```

2. *"UI `prepublishOnly` … runs the same guard on core."* — ui's `prepublishOnly` ends in `node ../../scripts/check-sdk-version-stamp.mjs ui`. The guard runs on **ui's own dist**, not core's.

This PR's own edit to `scripts/check-sdk-version-stamp.mjs` says the opposite: *"`@youversion/platform-react-ui` still bundles core through `noExternal`, so the same constant appears in its output until that bundling changes."*

**Concrete harm:** someone trusting this ADR concludes ui's dist carries no version constant, and drops ui from the guard list or from the `YVP_PUBLISH_BUILD` rebuild. ui then ships publicly with a `-dev` suffixed `X-YVP-Sdk` header and nothing catches it. The paragraph the PR replaced was correct; suggest restoring its substance.

packages/core/src/highlights.ts
```diff
@@ -76,23 +72,17 @@ export class HighlightsClient {
...
   private validateVersionId(value: number): void {
     try {
-      this.versionIdSchema.parse(value);
-    } catch (error) {
-      if (error instanceof z.ZodError) {
-        throw new Error('Version ID must be a positive integer');
-      }
-      throw error;
+      BibleVersionIdSchema.parse(value);
+    } catch {
```
Dustin-Kelley, line 76: **issue (non-blocking):** dropping the `instanceof z.ZodError` narrowing means the bare `catch {}` now relabels *every* throw from `.parse()` as the fixed validation string.

Base narrowed on `error instanceof z.ZodError` and rethrew anything else. Now any non-validation failure inside the schema — a `TypeError` from an undefined schema import after an entry-splitting regression (this PR introduces 17 tsup entries), or a throw from a future `.check()` refinement — surfaces to the caller as `"Version ID must be a positive integer"` with the real stack discarded. A caller passing a perfectly valid `versionId: 111` would see a bogus validation error and have no way to diagnose it.

Since validation failures now throw zod Mini's `$ZodError`, the narrowing can be restored as `error instanceof z.core.$ZodError` (rethrow otherwise). Same shape at `:84` and `:105`.

packages/core/scripts/minify-browser-css.mjs
```diff
@@ -0,0 +1,42 @@
...
+import { fileURLToPath } from 'node:url';
+
+const stylesDir = join(dirname(fileURLToPath(import.meta.url)), '..', 'src', 'styles');
+const outDir = join(dirname(fileURLToPath(import.meta.url)), '..', 'dist', 'styles');
+
+function minifyCss(css) {
+  return css
+    .replace(/\/\*[\s\S]*?\*\//g, '')
+    .replace(/\s+/g, ' ')
+    .replace(/\s*([{}:;,])\s*/g, '$1')
```
Dustin-Kelley, line 18: **issue (non-blocking):** `minifyCss` is string- and `url()`-unaware, so it can silently corrupt published CSS.

The rules run over the whole file with no tokenizer: `\s+ → ' '` and `\s*([{}:;,])\s* → '$1'`. Any of these in a source file would be rewritten *inside* the quoted value:

```css
content: "Chapter 1, verse 2";              /* space after the comma is eaten */
font-family: "YV Serif", serif;            /* ditto */
background: url("data:image/svg+xml;utf8,<svg viewBox=\"0 0 1 1\">");  /* URI broken */
```

Nothing catches it: the only post-minify checks are the non-empty guard plus four substring assertions (`--yv-reader-font-size`, `yv-bible-renderer`, `--yv-red`, `@import './preflight.css'`). This output is what core publishes under `./browser/styles/*` and what ui's Tailwind build imports, so a corrupted value ships to consumers.

Today's two source files happen to contain none of these constructs, so it's latent — it breaks the first time someone adds a quoted value. Worth either reaching for a real minifier (lightningcss is already in the Tailwind v4 dependency tree) or leaving these files unminified, since they're small and get minified again downstream by ui's `tailwindcss --minify`.

packages/core/package.json
```diff
@@ -9,11 +9,40 @@
...
     "dist",
-    "src/styles/*.css",
     "README.md",
     "CHANGELOG.md",
     "!dist/**/*.test.*",
-    "!dist/__tests__"
+    "!dist/__tests__",
+    "!dist/**/*.d.ts.map",
+    "!dist/**/*.d.cts",
+    "!dist/bible.js",
```
Dustin-Kelley, line 18: **suggestion (non-blocking):** the `files` denylist fails open.

Internal tsup entry outputs are excluded by 28 hand-enumerated `!dist/<entry>.js` / `!dist/<entry>.cjs` negations — one pair per entry — with nothing tying the list to `tsup.config.ts`.

Add an 18th entry (say `notes: 'src/notes.ts'`) and forget the two matching lines, and `npm publish` silently ships `dist/notes.js` and `dist/notes.cjs`. They're unreachable through the exports map (only `.`, `./browser`, `./browser/styles/*`, `./server` are public), so no test, typecheck, or guard fails — the only symptom is tarball bloat nobody notices until a size budget trips.

An allowlist (`dist/index.*`, `dist/browser.*`, `dist/server.*`, `dist/chunk-*`, `dist/styles/**`, `dist/**/*.d.ts`) would fail closed instead, and wouldn't need editing per entry.

packages/core/src/schemas/version.ts
```diff
@@ -31,3 +31,39 @@ export const BibleVersionSchema = z.object({
...
+  .optional(
+    z.object({
+      page_size: z.optional(z.union([z.int().check(z.positive()), z.literal('*')])),
+      page_token: z.optional(z.string()),
+      fields: z.optional(z.array(z.keyof(BibleVersionSchema))),
+      all_available: z.optional(z.boolean()),
+    }),
+  )
+  .check(
+    z.refine(
```
Dustin-Kelley, line 55: **nitpick (non-blocking):** this `z.refine` is unreachable, and the rule now lives in three places.

Both call sites re-implement the identical 1..3 field-count test on the raw `options` and throw a plain `Error` *before* `.parse()` runs — `packages/core/src/bible-versions.ts:52` and `packages/core/src/languages-list.ts:13-23` — so neither this refinement nor its twin at `packages/core/src/schemas/language.ts:78` can ever fire through the client.

The drift risk is concrete: change the bound to 1-5 here and behavior doesn't change at all, because the pre-throw still rejects at 4. Callers validating options through the exported schema directly get the refine; callers going through the client get the pre-throw.

If the pre-throw exists to preserve the plain-`Error` type (per the changeset's behavior note), a comment saying so would help — otherwise the refine is the better single source of truth.

Dustin-Kelley (commented):
_Posted by Claude, an AI agent, on behalf of Dustin._

Post-merge review pass over the full diff. Five findings below, using [conventionalcomments.org](https://conventionalcomments.org/) labels.

The high-risk mechanical claims in this PR were verified empirically in a worktree at the PR head rather than reasoned about, and all came back sound:

- **Nested `@import './preflight.css'` inside a `layer()`-imported stylesheet** — Tailwind v4.1.15 inlines it correctly; a full ui `build:css --minify` + `preserve-host-revert-layer.js` produces an 82,625-byte `dist/tailwind.css` containing the scoped preflight under `@layer yv-sdk-styles`.
- **`__SDK_PACKAGE_VERSION__` injection** — no vitest/vite/storybook config aliases core to `src`, so there is no `ReferenceError` path; `env` is still set in tsup so `process.env.YVP_PUBLISH_BUILD` still folds.
- **Stamp guard** — fails closed in *both* packages on clean dev builds (core: `chunk-5XY3YWRK.cjs`, `chunk-KRGGAERA.js`; ui: `index.cjs`, `index.js`; both exit 1). An apparent false-pass turned out to be a stale content-hashed chunk from running `tsup` directly and bypassing the `rm -rf dist` — the guard itself is sound.
- **Build ordering** — every turbo task is `dependsOn: ["^build"]`, so core's `dist/styles` always exists before ui's Tailwind run.
- **zod → zod/mini** — all schema conversions are semantics-preserving (notably `.optional().catch(undefined)` → `z.catch(z.optional(...), undefined)` and `.union().pipe()` → `z.pipe(z.union(...), ...)` preserve nesting order). `ZodError` has zero references in hooks/ui/examples. `country` going optional in `schemas/language.ts` is a fix — the base `countrySchema` was never wired up.
- **Cross-package types & tests** — core `tsc -p tsconfig.build.json` clean, hooks and ui `tsc --noEmit` clean against the new `.d.ts`, core 432/432 tests pass.

PR #387, opened 2026-09-08
ci: require human signoff before a major release

Kyleasmth:
Closes [YPE-2523](https://lifechurch.atlassian.net/browse/YPE-2523) — **pending a scope correction, see below.**

Ports the Swift SDK's major-release signoff gate ([YPE-2521](https://lifechurch.atlassian.net/browse/YPE-2521)) to this repo. A PR that declares a breaking change cannot merge until a write-access collaborator other than the author confirms it, by name, with the exact version and a 🚀.

## The ticket's stated mechanism does not exist here

YPE-2523 says *"`semantic-release` (current branch) detects a breaking change"*. This repo has no semantic-release: releases run on **Changesets**, and the bump level is declared in `.changeset/*.md` frontmatter rather than inferred from commit footers.

That is not a new discovery. `docs/release-hardening-decisions.md` (Decision 1, jhampton 2026-07-01) already records it:

> In the Swift SDK, conventional commits drive changelog + version math. **This repo does not work that way** … Version bump = the level declared in each changeset … No semantic-release / conventional-changelog tooling.

So the gate is the same; only the detector changes. `scripts/preview-release.mjs` keeps the Swift name and output contract, but asks `changeset status` instead of `@semantic-release/commit-analyzer`.

## One deliberate difference from Swift

Swift gates on "this release is a major". Here that would be wrong. Changesets accumulates changesets until the Version Packages PR ships them, so a major pending on `main` would make **every unrelated PR** report major and get blocked until the release went out.

The script therefore reports two separate things, and only the second gates:

| Field | Means |
| --- | --- |
| `is_major` | the pending release would be a major (informational) |
| `introduced_major` | **this PR** added a changeset declaring `major` (gates) |

## Verified behaviour

Run against real commits on this branch, not reasoned about:

| Scenario | Gate | `next` | `is_major` | `introduced_major` |
| --- | --- | --- | --- | --- |
| CI-only PR (empty changeset) | pass | — | false | false |
| Adds a `patch` changeset | pass | 2.12.2 | false | false |
| **Adds a `major` changeset** | **BLOCKS** | **3.0.0** | true | **true** |
| Unrelated patch, major already pending | pass | 3.0.0 | true | false |

The last row is the one a straight port would get wrong.

Also confirmed: the three packages are a `fixed` group in `.changeset/config.json`, so one major changeset bumps all three to the same `3.0.0`. That keeps Swift's single-version signoff format valid — worth knowing, because if `fixed` were ever removed the signoff comment would become ambiguous about which version is approved. The script fails loudly rather than picking one if it sees more than one version.

## Carried over from Swift unchanged

The parts that are release-tool-agnostic and were the hard-won bits there:

- Rejects **bot** comments and the **PR author**, so an author with write access cannot self-clear the gate.
- Verifies write/maintain/admin permission, with a per-login cache.
- Re-runs on comment `deleted`, so a cached `success` cannot outlive its evidence.
- Normalises `>` quote markers and whitespace, so GitHub's "Quote reply" matches.
- Upserts one blocking comment rather than posting a new one per push.
- Fails loudly if a major is detected but no version could be computed, rather than letting a degenerate regex pass the gate.

## Not done, needs someone with admin

The workflow only blocks if `major-release-signoff` is added as a **required status check** on `main`. I cannot read branch protection (404 without admin). This is what actually made the Swift gate bite, so it should not be skipped.

## Scope

Asking on the ticket for the description to be corrected before this merges, since it currently specifies a mechanism this repo does not have.

[YPE-2523]: https://lifechurch.atlassian.net/browse/YPE-2523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[YPE-2521]: https://lifechurch.atlassian.net/browse/YPE-2521?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

<h3>Greptile Summary</h3>

This PR adds a Changesets-aware workflow that requires an authorized collaborator to approve newly introduced major releases with the exact version and head commit.
- Computes release information and detects newly introduced major changesets.
- Separates untrusted preview execution from the status-writing gate for steady-state operation.
- Validates approver permissions and binds approvals to the full head SHA.
- Adds the parser dependency and an intentional empty changeset.

<h3>Confidence Score: 3/5</h3>

This PR is not yet safe to merge because its bootstrap run can still let pull-request-controlled release tooling produce the successful status intended to prove independent human approval.

The workflow restores trusted release tooling only when the detector already exists on main; during this initial rollout, it executes the pull request’s detector and lets that detector’s output control a write-capable success-status path without requiring collaborator signoff.

**Files Needing Attention:** .github/workflows/major-release-signoff.yml

| Filename | Overview |
|----------|----------|
| .github/workflows/major-release-signoff.yml | Adds the release-signoff workflow and hardens its steady-state trust boundary, but its bootstrap path still lets this pull request control the preview result used by the privileged gate. |
| scripts/preview-release.mjs | Computes pending Changesets versions and detects per-package major declarations introduced through additions, modifications, or renames. |
| package.json | Adds the pinned Changesets parser dependency used by the preview script. |
| pnpm-lock.yaml | Records the parser dependency and resulting peer-resolution updates. |
| .changeset/odd-towns-unite.md | Adds the required intentional empty changeset for this CI-only change. |

<h3>Flowchart</h3>

```mermaid
flowchart TD
  E[Pull request or comment event] --> C[Resolve current PR context]
  C --> F{Fork?}
  F -- Yes --> G[Fail closed and require signoff]
  F -- No --> P[Compute Changesets release preview]
  P --> M{PR introduced a major changeset?}
  M -- No --> S[Post success status]
  M -- Yes --> A[Search comments for exact version, full SHA, rocket, and authorized approver]
  A --> V{Valid signoff found?}
  V -- Yes --> S
  V -- No --> B[Post blocking comment and failure status]
```

<sub>Reviews (6): Last reviewed commit: ["Merge branch &#39;main&#39; into YPE-2523/major-..."](https://github.com/youversion/platform-sdk-react/commit/403a7a5baa79fd4b40ccf60fbabbeb45430203a8) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=62235045)</sub>

### DIFF
--- a/.github/workflows/major-release-signoff.yml
+++ b/.github/workflows/major-release-signoff.yml
@@ -157,8 +157,11 @@ jobs:
             *)    IS_MAJOR=0 ;;
           esac
           echo "is_major=$IS_MAJOR" >> "$GITHUB_OUTPUT"
-          echo "next=$(jq -r .next preview.json)" >> "$GITHUB_OUTPUT"
-          echo "release_type=$(jq -r .release_type preview.json)" >> "$GITHUB_OUTPUT"
+          # `// empty` rather than plain `-r`: jq prints a JSON null as the literal string
+          # "null", which is not empty, so the `${VAR:-default}` fallbacks downstream never
+          # fire and the status reads "No breaking changeset added (null)".
+          echo "next=$(jq -r '.next // empty' preview.json)" >> "$GITHUB_OUTPUT"
+          echo "release_type=$(jq -r '.release_type // empty' preview.json)" >> "$GITHUB_OUTPUT"
 
 
   gate:
--- a/.github/workflows/major-release-signoff.yml
+++ b/.github/workflows/major-release-signoff.yml
@@ -96,26 +96,26 @@ jobs:
         run: |
           set -euo pipefail
           git fetch --no-tags --force origin main:refs/heads/main
-          # Take main's copy where main has one. A PR that *introduces* one of these has
-          # nothing to restore, so it runs its own and the protection engages from the
-          # next PR onward. That is the bootstrap case, not a hole a PR can choose.
-          for f in scripts/preview-release.mjs package.json pnpm-lock.yaml \
-                   pnpm-workspace.yaml .changeset/config.json; do
-            if git cat-file -e "main:$f" 2>/dev/null; then
-              git checkout main -- "$f"
-            else
-              echo "::warning::$f is not on main yet; using this PR's copy."
-            fi
-          done
-          # Install hooks a PR could add to rewrite what we just restored. Take main's, or
-          # remove it outright: leaving a PR-authored one hands back the control.
-          for f in .npmrc .pnpmfile.cjs; do
-            if git cat-file -e "main:$f" 2>/dev/null; then
-              git checkout main -- "$f"
-            else
-              rm -f "$f"
-            fi
-          done
+          # All of main's tooling, or none of it. The detector, the dependency set it runs
+          # under, and the changesets config have to agree, so restoring some from main and
+          # taking the rest from the PR gives a detector without its own dependencies.
+          if git cat-file -e main:scripts/preview-release.mjs 2>/dev/null; then
+            for f in scripts/preview-release.mjs package.json pnpm-lock.yaml \
+                     pnpm-workspace.yaml .changeset/config.json; do
+              git cat-file -e "main:$f" 2>/dev/null && git checkout main -- "$f"
+            done
+            # Install hooks a PR could add to rewrite what we just restored. Take main's, or
+            # remove it outright: leaving a PR-authored one hands back the control.
+            for f in .npmrc .pnpmfile.cjs; do
+              if git cat-file -e "main:$f" 2>/dev/null; then
+                git checkout main -- "$f"
+              else
+                rm -f "$f"
+              fi
+            done
+          else
+            echo "::warning::scripts/preview-release.mjs is not on main yet, so this PR supplies its own tooling. The trust boundary engages once it merges."
+          fi
 
       - name: Setup pnpm
         uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # v4
--- a/.github/workflows/major-release-signoff.yml
+++ b/.github/workflows/major-release-signoff.yml
@@ -268,16 +268,17 @@ jobs:
           }
           NORM_TEMPLATE=$(normalize "$TEMPLATE_PHRASE")
 
-          # A signoff approves the code it was given on. Pushing more breaking changes after
-          # an approval must not inherit it, and it would: the next version is still the same
-          # string, so the old comment would keep matching. Ignore anything older than the
-          # commit currently under review.
-          HEAD_DATE=$(gh api "repos/${{ github.repository }}/commits/$HEAD_SHA" --jq '.commit.committer.date')
-          echo "Head $HEAD_SHA committed at $HEAD_DATE; older signoffs do not count."
+          # A signoff approves the code it was given on, so it names that code. Comparing
+          # timestamps instead would trust `committer.date`, which the committer sets, so a
+          # backdated commit could inherit an older approval. A SHA cannot be backdated, and
+          # a new push changes it, which retires the previous signoff on its own.
+          SHORT_SHA=${HEAD_SHA:0:7}
+          SHA_RE="(^|[^0-9A-Za-z])${SHORT_SHA}[0-9A-Fa-f]*([^0-9A-Za-z]|$)"
+          echo "Signoff must name head $SHORT_SHA."
 
           COMMENTS=$(gh api --paginate \
             "repos/${{ github.repository }}/issues/$PR_NUMBER/comments" \
-            --jq '[.[] | {login: .user.login, type: .user.type, body: .body, html_url: .html_url, at: (.updated_at // .created_at)}]')
+            --jq '[.[] | {login: .user.login, type: .user.type, body: .body, html_url: .html_url}]')
 
           declare -A PERM_CACHE
           MATCH_LOGIN=""
@@ -287,9 +288,6 @@ jobs:
             BODY=$(jq -r .body <<<"$row")
             URL=$(jq -r .html_url <<<"$row")
             USER_TYPE=$(jq -r .type <<<"$row")
-            AT=$(jq -r .at <<<"$row")
-            # Predates the commit under review, so it approved different code.
-            if [[ "$AT" < "$HEAD_DATE" ]]; then continue; fi
             # Bots can't sign off.
             if [ "$USER_TYPE" = "Bot" ]; then continue; fi
             # Skip the PR author — the point of the gate is a second pair of
@@ -299,6 +297,7 @@ jobs:
             NORM_BODY=$(normalize "$BODY")
             if [[ "$NORM_BODY" != *"$NORM_TEMPLATE"* ]]; then continue; fi
             if ! [[ "$BODY" =~ $VERSION_RE ]] || ! [[ "$BODY" =~ $ROCKET_RE ]]; then continue; fi
+            if ! [[ "$BODY" =~ $SHA_RE ]]; then continue; fi
             # Then verify write+ permission. Cached so a chatty approver doesn't
             # cost one API call per comment.
             PERM="${PERM_CACHE[$LOGIN]:-}"
@@ -342,9 +341,11 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           PR_NUMBER: ${{ needs.context.outputs.pr_number }}
+          HEAD_SHA: ${{ needs.context.outputs.head_sha }}
           NEXT_VERSION: ${{ steps.decision.outputs.next }}
         run: |
           set -euo pipefail
+          SHORT_SHA=${HEAD_SHA:0:7}
           MARKER='<!-- major-signoff-required -->'
           # echo-per-line rather than a heredoc: un-indented heredoc lines
           # confuse the YAML block-scalar parser.
@@ -352,11 +353,12 @@ jobs:
             echo "$MARKER"
             echo "## 🚨 Breaking change detected — signoff required for \`v$NEXT_VERSION\`"
             echo
-            echo "This PR adds a changeset declaring a **major** bump, which would ship all three packages as \`v$NEXT_VERSION\` (they are a \`fixed\` group in \`.changeset/config.json\`, so they version together). Merging is blocked until a repo collaborator with write access comments on this PR with **all three (3) of the following:**"
+            echo "This PR adds a changeset declaring a **major** bump, which would ship all three packages as \`v$NEXT_VERSION\` (they are a \`fixed\` group in \`.changeset/config.json\`, so they version together). Merging is blocked until a repo collaborator with write access comments on this PR with **all four (4) of the following:**"
             echo
             echo "1. The verbatim acknowledgment phrase below,"
-            echo "2. The precise next version (\`v$NEXT_VERSION\` or \`$NEXT_VERSION\`), and"
-            echo "3. A 🚀 (\`:rocket:\`) emoji."
+            echo "2. The precise next version (\`v$NEXT_VERSION\` or \`$NEXT_VERSION\`),"
+            echo "3. The commit being approved (\`$SHORT_SHA\`), and"
+            echo "4. A 🚀 (\`:rocket:\`) emoji."
             echo
             echo "I confirm that this is an intentional breaking change, and I have read the release procedures. I understand and have documented its impact upon release."
             echo
@@ -365,10 +367,10 @@ jobs:
             echo '```'
             echo "I confirm that this is an intentional breaking change, and I have read the release procedures. I understand and have documented its impact upon release."
             echo
-            echo "v$NEXT_VERSION 🚀"
+            echo "v$NEXT_VERSION $SHORT_SHA 🚀"
             echo '```'
             echo
-            echo "The check re-runs automatically when a qualifying comment is posted or edited."
+            echo "The check re-runs automatically when a qualifying comment is posted or edited. Pushing new commits changes the hash, so a fresh signoff is needed."
           } > comment.md
 
           EXISTING=$(gh api --paginate \
--- a/.github/workflows/major-release-signoff.yml
+++ b/.github/workflows/major-release-signoff.yml
@@ -8,10 +8,10 @@ on:
     # so a previously-cached `success` status doesn't outlive its evidence.
     types: [created, edited, deleted]
 
-permissions:
-  contents: read
-  pull-requests: write
-  statuses: write
+
+# Declared per job. The job that runs PR-authored code holds no write scopes, and the
+# job that can write statuses and comments never checks that code out.
+permissions: {}
 
 concurrency:
   group: major-release-signoff-${{ github.event.pull_request.number || github.event.issue.number }}
@@ -21,12 +21,19 @@ env:
   STATUS_CONTEXT: major-release-signoff
 
 jobs:
-  signoff:
+  context:
     # issue_comment fires for issues AND PRs across the repo. Skip non-PR
     # issues so we don't waste a run when someone comments on a plain issue.
     if: ${{ github.event_name != 'issue_comment' || github.event.issue.pull_request != null }}
-    name: Major release signoff
+    name: Resolve PR context
     runs-on: ubuntu-latest
+    permissions:
+      pull-requests: read
+    outputs:
+      head_sha: ${{ steps.pr.outputs.head_sha }}
+      is_fork: ${{ steps.pr.outputs.is_fork }}
+      pr_author: ${{ steps.pr.outputs.pr_author }}
+      pr_number: ${{ steps.pr.outputs.pr_number }}
     steps:
       - name: Resolve PR context
         id: pr
@@ -38,6 +45,7 @@ jobs:
             PR_NUMBER=${{ github.event.pull_request.number }}
             HEAD_SHA=${{ github.event.pull_request.head.sha }}
             PR_AUTHOR=${{ github.event.pull_request.user.login }}
+            HEAD_REPO=${{ github.event.pull_request.head.repo.full_name }}
           else
             PR_NUMBER=${{ github.event.issue.number }}
             # issue_comment payload has no head/base info — look it up so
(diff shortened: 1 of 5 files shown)


Kyleasmth:
@jhampton @camrun91 — out of draft and green. Two different asks, so splitting them:

**@jhampton** — the open question is on [YPE-2523](https://lifechurch.atlassian.net/browse/YPE-2523), not in this diff. The ticket specifies `semantic-release` detecting the breaking change; this repo has none, so the detector here is `changeset status` instead. That is not a new finding — your Decision 1 on YPE-2486 already records that this repo does not work the Swift way. The code is correct either way; the ticket wording is what needs your call.

The one thing that would change this diff: I gate on **"this PR added a `major` changeset"**, not "the pending release is a major". Gating on the latter would block every unrelated PR while a major sits pending on `main`. If you want the Version Packages PR gated as a second line of defence, that is a different detector and I would rather add it deliberately than assume it.

**@camrun91** — normal review welcome, particularly on whether the workflow fits how this repo runs CI.

**Worth knowing, since it is the interesting bit:** the gate failed on its own PR when I first pushed it, and the reason was hidden because the script swallowed stderr. `changeset status` resolves the configured `baseBranch` as a **local** ref, and a PR checkout only has `origin/main` — it worked locally purely because my clone had a real `main` branch. Fixed by fetching `main:refs/heads/main` in the workflow, and the script now surfaces Changesets' own message rather than exiting silently. It is passing its own gate now, which is a real test rather than a synthetic one: this PR carries an empty changeset, so `introduced_major` is false and the check goes green on its own merits.

**Still needs someone with admin:** the gate only actually blocks once `major-release-signoff` is a required status check on `main`. I get a 404 reading branch protection.


[YPE-2523]: https://lifechurch.atlassian.net/browse/YPE-2523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

camrun91:
> @jhampton @camrun91 — out of draft and green. Two different asks, so splitting them:
> 
> **@jhampton** — the open question is on [YPE-2523](https://lifechurch.atlassian.net/browse/YPE-2523), not in this diff. The ticket specifies `semantic-release` detecting the breaking change; this repo has none, so the detector here is `changeset status` instead. That is not a new finding — your Decision 1 on YPE-2486 already records that this repo does not work the Swift way. The code is correct either way; the ticket wording is what needs your call.
> 
> The one thing that would change this diff: I gate on **"this PR added a `major` changeset"**, not "the pending release is a major". Gating on the latter would block every unrelated PR while a major sits pending on `main`. If you want the Version Packages PR gated as a second line of defence, that is a different detector and I would rather add it deliberately than assume it.
> 
> **@camrun91** — normal review welcome, particularly on whether the workflow fits how this repo runs CI.
> 
> **Worth knowing, since it is the interesting bit:** the gate failed on its own PR when I first pushed it, and the reason was hidden because the script swallowed stderr. `changeset status` resolves the configured `baseBranch` as a **local** ref, and a PR checkout only has `origin/main` — it worked locally purely because my clone had a real `main` branch. Fixed by fetching `main:refs/heads/main` in the workflow, and the script now surfaces Changesets' own message rather than exiting silently. It is passing its own gate now, which is a real test rather than a synthetic one: this PR carries an empty changeset, so `introduced_major` is false and the check goes green on its own merits.
> 
> **Still needs someone with admin:** the gate only actually blocks once `major-release-signoff` is a required status check on `main`. I get a 404 reading branch protection.

Did you mean to tag @cameronapak? 

[YPE-2523]: https://lifechurch.atlassian.net/browse/YPE-2523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

.github/workflows/major-release-signoff.yml
```diff
@@ -0,0 +1,278 @@
...
+        uses: pnpm/action-setup@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: 24
+          cache: 'pnpm'
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
```
cameronapak: **[issue](https://conventionalcomments.org/): Do not install PR-controlled code in the write-capable signoff job.**

<details>
<summary>For Agents: untrusted PR execution</summary>

The `issue_comment` path checks out the current PR head and runs `pnpm install` while this job has `pull-requests: write` and `statuses: write`. A fork can change dependency metadata or lifecycle scripts, then trigger this path with a comment; that code runs before the signoff decision and can forge the success status or alter PR comments. Keep PR analysis in a read-only job, and run status/comment mutation only from trusted workflow code without installing the PR head.

</details>

By Code Reviewer bot, sent on behalf of Cam.

scripts/preview-release.mjs
```diff
@@ -0,0 +1,138 @@
...
+    rmSync(dir, { recursive: true, force: true });
+  }
+}
+
+/** Bump levels declared by the changeset files this PR adds. */
+function addedChangesetLevels(base) {
+  const added = git(
+    'diff',
+    '--name-only',
+    '--diff-filter=A',
```
cameronapak: **[issue](https://conventionalcomments.org/): Include modified changesets when detecting a major bump.**

<details>
<summary>For Agents: changeset diff filter</summary>

`--diff-filter=A` only examines newly added `.changeset/*.md` files. A PR can modify an existing unconsumed patch/minor changeset to declare `major`; `changeset status` will report a major release, but `introduced_major` remains false and the workflow posts a successful “signoff not required” status. Inspect added and modified changesets, using the final PR contents.

</details>

By Code Reviewer bot, sent on behalf of Cam.

.github/workflows/major-release-signoff.yml
```diff
@@ -0,0 +1,278 @@
...
+            PR_JSON=$(gh api repos/${{ github.repository }}/pulls/$PR_NUMBER)
+            HEAD_SHA=$(jq -r .head.sha <<<"$PR_JSON")
+            PR_AUTHOR=$(jq -r .user.login <<<"$PR_JSON")
+          fi
+          echo "pr_number=$PR_NUMBER" >> "$GITHUB_OUTPUT"
+          echo "head_sha=$HEAD_SHA" >> "$GITHUB_OUTPUT"
+          echo "pr_author=$PR_AUTHOR" >> "$GITHUB_OUTPUT"
+
+      - name: Checkout PR head
+        uses: actions/checkout@v5
```
cameronapak: **[issue](https://conventionalcomments.org/): Pin the actions used by this privileged workflow to full commit SHAs.**

<details>
<summary>For Agents: mutable action references</summary>

`actions/checkout@v5`, `pnpm/action-setup@v4`, and `actions/setup-node@v6` are mutable tag references inside a workflow that can write statuses and PR comments. Pin each to a reviewed full commit SHA, consistent with the pinned privileged actions already used in `.github/workflows/release.yml`.

</details>

By Code Reviewer bot, sent on behalf of Cam.

cameronapak (commented):
## Review

[Jira: YPE-2523](https://lifechurch.atlassian.net/browse/YPE-2523). Spec is the PR body.

## Summary

Standards: 2 must-fix findings. Spec: 2 must-fix findings. Worst: the comment-triggered path executes PR-controlled install code with write-capable credentials.

<details>
<summary>For Agents: review basis</summary>

- CI line: `major-release-signoff`
- Greptile/bot: prior findings were independently verified.
- Event: `issue_comment`
- HEAD: `fb94946456e826775b6fca1d75aeb2f50d94d829`

</details>

By Code Reviewer bot, sent on behalf of Cam.

cameronapak:
**[suggestion](https://conventionalcomments.org/): Prefer GitHub-native approvals over a custom comment-driven release gate.**

<details>
<summary>For Agents: simpler release control</summary>

A simpler approach would be to assign release maintainers as CODEOWNERS for `.changeset/*`, require code-owner approval on `main`, and dismiss stale approvals when new commits are pushed. This makes every release-bearing PR—including a major changeset—require explicit review from someone responsible for release impact, while automatically covering modified changesets.

That removes the `issue_comment` workflow, exact phrase/version/emoji parsing, PR-head installation, custom status management, and privileged mutable action execution. If approval must happen only at publication time, a protected GitHub Environment with required reviewers on the npm publish job is the next-simplest native option.

The tradeoff is that CODEOWNERS requires approval for every release-bearing changeset rather than only majors, but it directly protects the stated goal with substantially less custom security-sensitive code.

</details>

By Code Reviewer bot, sent on behalf of Cam.

Kyleasmth:
> > @jhampton @camrun91 — out of draft and green. Two different asks, so splitting them:
> > **@jhampton** — the open question is on [YPE-2523](https://lifechurch.atlassian.net/browse/YPE-2523), not in this diff. The ticket specifies `semantic-release` detecting the breaking change; this repo has none, so the detector here is `changeset status` instead. That is not a new finding — your Decision 1 on YPE-2486 already records that this repo does not work the Swift way. The code is correct either way; the ticket wording is what needs your call.
> > The one thing that would change this diff: I gate on **"this PR added a `major` changeset"**, not "the pending release is a major". Gating on the latter would block every unrelated PR while a major sits pending on `main`. If you want the Version Packages PR gated as a second line of defence, that is a different detector and I would rather add it deliberately than assume it.
> > **@camrun91** — normal review welcome, particularly on whether the workflow fits how this repo runs CI.
> > **Worth knowing, since it is the interesting bit:** the gate failed on its own PR when I first pushed it, and the reason was hidden because the script swallowed stderr. `changeset status` resolves the configured `baseBranch` as a **local** ref, and a PR checkout only has `origin/main` — it worked locally purely because my clone had a real `main` branch. Fixed by fetching `main:refs/heads/main` in the workflow, and the script now surfaces Changesets' own message rather than exiting silently. It is passing its own gate now, which is a real test rather than a synthetic one: this PR carries an empty changeset, so `introduced_major` is false and the check goes green on its own merits.
> > **Still needs someone with admin:** the gate only actually blocks once `major-release-signoff` is a required status check on `main`. I get a 404 reading branch protection.
> 
> Did you mean to tag @cameronapak?

Yes, thank you.

[YPE-2523]: https://lifechurch.atlassian.net/browse/YPE-2523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Kyleasmth:
All three addressed in `5ba79fc` and `c5fac06`. Thanks both — chasing these turned up more than the original three, so worth writing down what changed.

## Privileged workflow executing PR code

Split into three jobs. The one that runs PR code holds no write scopes; the one that can write statuses and comments never checks that code out.

| Job | Permissions | Runs PR code |
| --- | --- | --- |
| `context` | `pull-requests: read` | no |
| `preview` | `contents: read` | yes |
| `gate` | `pull-requests: write`, `statuses: write` | no |

Workflow-level is now `permissions: {}`. Fork PRs skip `preview` entirely and the gate fails closed rather than trusting a preview computed by untrusted code.

That alone was not enough, though. A same-repo author could still edit `scripts/preview-release.mjs` to return `introduced_major: false` and the trusted job would post green. So the preview job now restores the detector from `main` before anything installs: `scripts/preview-release.mjs`, `package.json`, `pnpm-lock.yaml`, `pnpm-workspace.yaml`, `.changeset/config.json`, and `.npmrc` / `.pnpmfile.cjs` restored-or-deleted. The PR supplies changeset **data** only.

`--ignore-scripts` also was not sufficient on its own: pnpm hooks are a separate mechanism, so a PR-authored `.pnpmfile.cjs` could rewrite the detector after restore. Install is now `--ignore-scripts --ignore-pnpmfile`.

## Modified changesets bypassing the gate

Confirmed exactly as described, against real commits. Editing a pending changeset from `patch` to `major` gave:

| | `introduced_major` | `is_major` |
| --- | --- | --- |
| before | **false** | true |
| after | **true** | true |

So the workflow posted "No breaking changeset added; signoff not required" while Changesets computed a real major.

Now `--diff-filter=AMR` with rename following, and the base comparison is **per package** so an existing major on one package cannot mask a different one being raised. Three further bypasses fell out while testing, each proved with real commits:

* **Renaming** a changeset while raising it to major reported as `R` and was skipped entirely.
* **Filenames git C-quotes** (a literal tab, say) failed the changeset name test and dropped out of the scan. Diff parsing is now `-z` NUL-delimited.
* **YAML forms the line regex missed**: `{"pkg": major}`, anchors/aliases, block scalars, `!!str`, quoted levels. The hand-written parser is gone; `@changesets/parse` is a real dependency now, pinned to `0.4.1`, the exact version CLI 2.29.7 resolves, so it cannot drift from what `changeset status` itself computes.

## Mutable action tags

All three pinned to full SHAs with the tag in a trailing comment.

## One more, not in the original review

A signoff was not bound to the head it was given on. Approve `v3.0.0` on commit A, push commit B with *additional* breaking changes, version is still `3.0.0`, and the old comment satisfied the gate. This one can happen by accident rather than by malice. Comments older than the commit under review no longer count.

Also fixed: on a preview failure the job used to abort before reaching the step that posts a red status, so a previously-green status could outlive the signoff that earned it. The decision step now records why it cannot evaluate and the red-status and fail steps run under `always()`.

## Known residuals

* If the `context` lookup itself fails, `gate` is skipped and a stale green can remain on the SHA. Cannot be fixed by the job condition alone, since an `issue_comment` payload carries no head SHA. The mitigation would be setting the status to `pending` in the trusted phase first; happy to add it if you want it.
* A same-repo collaborator can still edit the workflow file itself. That is inherent to the gate living in the repo it guards.

.github/workflows/major-release-signoff.yml
Kyleasmth: Done. The job that runs PR code now holds no write scopes, and the one that posts statuses never checks that code out. Install is also `--ignore-scripts --ignore-pnpmfile`, and the detector plus its lockfile are restored from `main` so a branch cannot rewrite what gates it.

scripts/preview-release.mjs
Kyleasmth: Done, and it was real: I reproduced it on commits. Editing a pending changeset from `patch` to `major` gave `introduced_major=false` while `is_major=true`, so the gate posted "signoff not required" for a genuine major. Now `--diff-filter=AMR` with rename following, compared per package.

.github/workflows/major-release-signoff.yml
Kyleasmth: Done. All three pinned to full commit SHAs with the tag kept in a trailing comment.

.github/workflows/major-release-signoff.yml
```diff
@@ -0,0 +1,410 @@
...
+            printf '%s' "$1" | tr -d '>' | tr -s '[:space:]' ' '
+          }
+          NORM_TEMPLATE=$(normalize "$TEMPLATE_PHRASE")
+
+          # A signoff approves the code it was given on, so it names that code. Comparing
+          # timestamps instead would trust `committer.date`, which the committer sets, so a
+          # backdated commit could inherit an older approval. A SHA cannot be backdated, and
+          # a new push changes it, which retires the previous signoff on its own.
+          SHORT_SHA=${HEAD_SHA:0:7}
+          SHA_RE="(^|[^0-9A-Za-z])${SHORT_SHA}[0-9A-Fa-f]*([^0-9A-Za-z]|$)"
```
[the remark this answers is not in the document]
Kyleasmth: Fixed in `9e25007`. The matcher now requires the full 40-character SHA, and the copy-paste template hands the reviewer the full hash. A 7-character prefix is 28 bits, which is cheap enough to grind, so this was worth closing.

.github/workflows/major-release-signoff.yml
```diff
@@ -0,0 +1,411 @@
...
+  cancel-in-progress: true
+
+env:
+  STATUS_CONTEXT: major-release-signoff
+
+jobs:
+  context:
+    # issue_comment fires for issues AND PRs across the repo. Skip non-PR
+    # issues so we don't waste a run when someone comments on a plain issue.
+    if: ${{ github.event_name != 'issue_comment' || github.event.issue.pull_request != null }}
```
jhampton, line 27: Nice

.github/workflows/major-release-signoff.yml
```diff
@@ -0,0 +1,411 @@
...
+    name: Resolve PR context
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: read
+    outputs:
+      head_sha: ${{ steps.pr.outputs.head_sha }}
+      is_fork: ${{ steps.pr.outputs.is_fork }}
```
jhampton, line 34: Good thinking.

.github/workflows/major-release-signoff.yml
```diff
@@ -0,0 +1,411 @@
...
+          # "Failed to find where HEAD diverged from main" if it is missing. A PR checkout
+          # has only origin/main, so create the local branch too.
+          git fetch --no-tags --force origin main:refs/heads/main
+          # Merge-base, not the PR's base sha: on a synchronize event the base
+          # branch may have moved, and diffing against its tip would attribute
+          # someone else's changesets to this PR.
+          BASE=$(git merge-base main "$HEAD_SHA")
+          node scripts/preview-release.mjs --base "$BASE" --head "$HEAD_SHA" > preview.json
+          cat preview.json
+          # `introduced_major`, not `is_major`: a major already pending on main
```
jhampton, line 151: NOICE

.github/workflows/major-release-signoff.yml
```diff
@@ -0,0 +1,411 @@
...
+      next: ${{ steps.preview.outputs.next }}
+      release_type: ${{ steps.preview.outputs.release_type }}
+    steps:
+      - name: Checkout PR head
+        uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5
+        with:
+          ref: ${{ needs.context.outputs.head_sha }}
+          fetch-depth: 0
+
+      - name: Restore release tooling from the base branch
```
jhampton, line 91: Security, nice.

.github/workflows/major-release-signoff.yml
```diff
@@ -0,0 +1,411 @@
...
+
+  gate:
+    # `always()`: `preview` is skipped for forks and the gate still has to post a status.
+    if: always() && needs.context.result == 'success'
+    name: Major release signoff
+    needs: [context, preview]
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+      statuses: write
```
jhampton, line 175: I get this, but I'm not sure it's something we _must_ guard against.  That said, it could happen accidentally, so we need _some_ kind of answer.  How about we check the file modifications against the workflows folder and establish a rule that breaking change PRs cannot introduce or change CI behavior, i.e. That would need to be a separate PR. What do you think?

Kyleasmth: Worth having, but one catch: on `pull_request` GitHub runs the workflow file from the PR head, so a PR could delete the new rule in the same commit. Any guard in this file is editable by the PR it guards.

What a PR cannot touch is on your side:
- make `major-release-signoff` a required check
- "Require review from Code Owners" + a `CODEOWNERS` entry for `.github/workflows/**`

No CODEOWNERS in the repo today. Who should own that path?

jhampton: Ah CODEOWNERS....I know thee well.  Not just yet, but I'd entertain a sweep across all SDKs in the future with a proper OWNERS or CODEOWNERS structure.

The required check has been added.

jhampton (approved):
Image

The required check has been added, and that is sufficient for me to approve the PR even with the Greptile P1 security finding. Because the check is in fact required, any PR that removes that check would block a merge to `main`. So at this point, I'm feeling pretty good about it.

Nicely done, thank 'ya.

PR #393, opened 2026-09-10
feat(core): add passage display model

cameronapak:
## Summary

- add a high-level getPassageDisplay API for transformed Bible HTML, current attribution, stylesheet descriptors, and required container attributes
- preserve a declarative core boundary with no DOM mutation, resource injection, or attribution caching
- document the contract and decision, expose tree-shakable helpers, and add a core minor changeset

## Behavior

- always returns transformed HTML
- fetches passage and version metadata concurrently when filters permit
- prefers copyright attribution, falls back to promotional content, and fails closed when neither exists
- provides the versioned Bible CSS and Fonts API stylesheet for permanent Untitled Serif font ID 1

## Verification

- pnpm lint
- YVP_API_HOST=api.youversion.com pnpm --filter @youversion/platform-core test (22 files, 442 tests)
- pnpm check:tree-shaking
- pnpm --filter @youversion/platform-core build

<h3>Greptile Summary</h3>

Adds a declarative core API that combines transformed passage HTML, current attribution, stylesheet descriptors, and required container attributes without mutating the DOM.
- Exposes facade and tree-shakable passage-display APIs with schema-derived public types.
- Preserves version-filter enforcement while coordinating passage and metadata requests.
- Adds CSS-major and tree-shaking checks, focused tests, documentation, and a core changeset.

The pull request appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/core/src/bible-passage-display.ts | Implements the declarative display-model orchestration, attribution selection, filtering behavior, and concurrent request path. |
| packages/core/src/bible-passage.ts | Extracts reusable query, fetch, and transformation helpers while adding schema validation to the prevalidated-version path. |
| packages/core/src/schemas/passage-display.ts | Defines runtime schemas and readonly public types for display options, attribution, stylesheets, and the complete model. |
| packages/core/src/__tests__/bible-passage-display.test.ts | Explicitly installs display-test setup within each HTTP-capable test, resolving the previously reported hidden suite-level setup concern. |
| packages/core/src/bible-display-resources.ts | Provides stable Bible and font stylesheet descriptors using validated app configuration. |
| scripts/check-cdn-css-major.mjs | Adds a repository check keeping the core stylesheet URL aligned with the UI CDN compatibility major. |

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
  participant App as Consumer
  participant Display as getPassageDisplay
  participant API as Platform API
  participant Transform as HTML transformer
  App->>Display: versionId, passageId, options
  alt Language filter requires metadata
    Display->>API: Fetch and validate version
    API-->>Display: BibleVersion
    Display->>API: Fetch HTML passage
  else Version ID can be checked locally
    par Concurrent requests
      Display->>API: Fetch BibleVersion
      Display->>API: Fetch HTML passage
    end
  end
  API-->>Display: Passage HTML
  Display->>Transform: Transform HTML
  Transform-->>Display: Reader-ready HTML
  Display-->>App: Display model, attribution, stylesheets, attributes
```

<sub>Reviews (6): Last reviewed commit: ["fix(core): preserve getPassage response ..."](https://github.com/youversion/platform-sdk-react/commit/5ef2001432b82548d99bac50affc7ba30e7551f4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=62577792)</sub>

<details><summary><h4>Context used (4)</h4></summary>

- Knowledge Base — [Core platform SDK](https://app.greptile.com/youversion/-/custom-context/knowledge-base/youversion/platform-sdk-react/-/docs/core-platform.md)
- Knowledge Base — [Core API clients and data contracts](https://app.greptile.com/youversion/-/custom-context/knowledge-base/youversion/platform-sdk-react/-/docs/core-api-data.md)
- Knowledge Base — [Bible content services](https://app.greptile.com/youversion/-/custom-context/knowledge-base/youversion/platform-sdk-react/-/docs/core-bible-content.md)
- Knowledge Base — [Monorepo build, validation, and release](https://app.greptile.com/youversion/-/custom-context/knowledge-base/youversion/platform-sdk-react/-/docs/monorepo-build-release.md)
</details>


packages/core/src/bible-passage-display.ts
```diff
@@ -0,0 +1,126 @@
...
+import { BibleVersionSchema } from './schemas/version';
+import type { ApiConfig, BibleVersion } from './types';
+import {
+  isLanguageFilterActive,
+  isVersionIdDecidablyUnusable,
+  throwUnusableBibleVersion,
+} from './version-filters';
+
+/** The versioned YouVersion stylesheet for Bible HTML. */
+export const BIBLE_CSS_STYLESHEET_URL = 'https://cdn.youversion.com/platform/1/bible.css';
```
cameronapak: question: How does versioning work for the Bible stylesheet CSS? 

Is it that the version follows the version of the core package??

packages/core/src/bible-passage-display.ts
abharms: issue (blocking): keep the stylesheet resource on one synchronized, narrow bundle path

Two release-contract problems meet here:

- `PUBLISHING.md` names `packages/ui/CDN_CSS_MAJOR_VERSION` as the CSS compatibility-major authority, but this module independently hardcodes `/platform/1/`. A future major bump can leave newly published core builds silently returning stale CSS. Please add a mandatory repository/CI check that asserts this URL's major matches the publisher's version.
- A production consumer importing only `getBibleStylesheets` currently retains this module's Zod schemas and passage orchestration (about 30.4 KB minified / 10.0 KB gzip in a direct bundle check). Please move the constants/helper into a dependency-free module and add a helper-only case to `scripts/check-tree-shaking.mjs`.

_Cursor, an AI agent sent on behalf of Austin._

packages/core/src/__tests__/bible-passage-display.test.ts
abharms: issue (blocking): make this suite exercise the promised contract without escaping to live traffic

Two gaps leave important behavior unprotected:

- These cases use `server.use()` with a fake app key, but shared setup does not start MSW when `INTEGRATION_TESTS=true`, while `test:integration` runs all Vitest files. The handlers then cannot intercept, so this mocked suite may send unintended requests to the configured API host. Please keep MSW running for this file in integration mode, as `client.test.ts` does, or skip this suite for live runs.
- The "current attribution" test makes only one display request, so it still passes if a later implementation caches the first attribution forever. Please make two calls, change the version metadata response between them, and assert the second result uses the updated attribution.

_Cursor, an AI agent sent on behalf of Austin._

CONTEXT.md
abharms: question (blocking): is fail-closed display attribution scoped to this model or to every SDK-rendered passage?

This definition reads as project-wide, but the existing `BibleReader` and `BibleCard` still render passages with copyright only: neither falls back to `promotional_content`, and both can display content without attribution. The new core operation instead rejects when neither field exists.

Please make the intended scope explicit. If this is SDK-wide legal policy, align the UI surfaces or link a concrete tracked follow-up; if it is specific to `BiblePassageDisplay`, narrow this definition and the API documentation accordingly.

_Cursor, an AI agent sent on behalf of Austin._

packages/core/src/bible-passage.ts
abharms: issue (non-blocking): validate public arguments before version-filter I/O

Moving `assertUsableVersion` ahead of boolean parsing changes established behavior for malformed JavaScript or unsafe-TypeScript calls. With a language filter, an invalid `include_headings` now causes a metadata request before the Zod error; with an excluded version, the same call returns `403` instead of the previous validation error.

Please validate all public arguments first, then enforce the version filter and call the already-validated fetch path.

_Cursor, an AI agent sent on behalf of Austin._

packages/core/src/bible-passage-display.ts
```diff
@@ -0,0 +1,129 @@
...
+  isLanguageFilterActive,
+  isVersionIdDecidablyUnusable,
+  throwUnusableBibleVersion,
+} from './version-filters';
+
+/**
+ * The YouVersion stylesheet for Bible HTML. The path is a CSS compatibility
+ * major, not the version of this package; routine UI releases overwrite it.
+ */
+export const BIBLE_CSS_STYLESHEET_URL = 'https://cdn.youversion.com/platform/1/bible.css';
```
abharms, line 25: issue (blocking): keep this compatibility major synchronized with the CDN publisher

`PUBLISHING.md` names `packages/ui/CDN_CSS_MAJOR_VERSION` as the authority, but this independently hardcodes `1`. After a breaking CSS bump, newly published core builds could silently keep returning stale CSS.

Suggested change: add a required check (similar to `check-sdk-version-stamp.mjs`) that parses this URL and asserts its path major equals the trimmed contents of `packages/ui/CDN_CSS_MAJOR_VERSION`. Wire it into the root validation/CI command so the two values cannot drift.

_Cursor, an AI agent sent on behalf of Austin._

packages/core/src/bible-passage-display.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  constructor(versionId: number) {
+    super(`Bible version ${versionId} has no display attribution.`);
+    this.name = 'MissingPassageAttributionError';
+    this.versionId = versionId;
+  }
+}
+
+/** Returns the ordered stylesheet resources for displaying Bible HTML. */
+export function getBibleStylesheets(
```
abharms, line 48: suggestion (non-blocking): keep this resource-only helper on a narrow bundle path

A production bundle importing only `getBibleStylesheets` still retains this module's Zod schemas and passage orchestration (30,391 bytes minified / 10,037 bytes gzip in a direct bundle check). For a resource-only helper, that is a disproportionately large dependency footprint.

Suggested change: move `BIBLE_CSS_STYLESHEET_URL`, `UNTITLED_SERIF_FONT_ID`, and `getBibleStylesheets` into a dependency-free `bible-display-resources.ts`; import that module from here and re-export it from the public entry point. Add a helper-only fixture to `scripts/check-tree-shaking.mjs` with a metafile or sentinel assertion that Zod and display orchestration do not enter this graph, optionally backed by a generous size ceiling.

_Cursor, an AI agent sent on behalf of Austin._

packages/core/src/__tests__/bible-passage-display.test.ts
```diff
@@ -0,0 +1,253 @@
...
+
+function clearVersionFilters(): void {
+  YouVersionPlatformConfiguration.permittedVersionIds = undefined;
+  YouVersionPlatformConfiguration.excludedVersionIds = undefined;
+  YouVersionPlatformConfiguration.permittedLanguageTags = undefined;
+}
+
+function setupDisplayTest(): void {
+  clearVersionFilters();
+  server.use(
```
abharms, line 41: issue (blocking): prevent this mocked suite from making live integration requests

`test:integration` runs all Vitest files, but shared setup does not start MSW when `INTEGRATION_TESTS=true`. In that mode this `server.use()` handler cannot intercept, and the fake app key may be sent to the configured API host.

Suggested change: when `INTEGRATION_TESTS` is set, conditionally mirror the explicit `beforeAll(() => server.listen({ onUnhandledRequest: 'error' }))` / `afterAll(() => server.close())` lifecycle used by `client.test.ts`, or gate this entire mocked suite with `describe.skipIf(process.env.INTEGRATION_TESTS === 'true')`.

_Cursor, an AI agent sent on behalf of Austin._

packages/core/src/__tests__/bible-passage-display.test.ts
```diff
@@ -0,0 +1,253 @@
...
+    http.get(`https://${apiHost}/v1/bibles/:id`, () => HttpResponse.json(mockDisplayVersion)),
+  );
+}
+
+describe('passage display model', () => {
+  it('returns transformed HTML, current attribution, stylesheets, and container attributes', async () => {
```
abharms, line 47: suggestion (non-blocking): directly test the promised no-cache attribution behavior

This makes only one display request, so it would still pass if a future implementation added SDK-level memoization and cached the first attribution indefinitely. The current implementation is stateless; this is a regression-coverage gap for the documented guarantee.

Suggested change: have the version handler return `copyright: 'First'` on the first request and `copyright: 'Second'` on the next; call `getPassageDisplay` twice with the same client/options and assert the second result contains `'Second'`. Also assert the version endpoint was called twice.

_Cursor, an AI agent sent on behalf of Austin._

CONTEXT.md
```diff
@@ -20,6 +20,23 @@ A verse or contiguous verse range in one chapter of one Bible version,
...
+
+## Passage display model
+
+A declarative representation of a passage and everything a web application
+needs to display it with YouVersion's rendering contract. _Avoid_: rendered
+passage, passage component, display bundle.
+
+## Display attribution
+
+The current, non-empty legal text that accompanies a displayed passage. The
```
abharms, line 36: question (blocking): define whether this contract applies to every SDK-rendered passage

As written, this is project-wide vocabulary, but `BibleReader` and `BibleCard` still render passages with copyright only: neither falls back to `promotional_content`, and both can display content without attribution. The new core model instead fails closed.

Suggested change: choose and document one contract. If this is SDK-wide, update both UI components to use the copyright → promotional-content fallback and define their explicit missing-attribution behavior; a shared core selector is one implementation option. If it is model-only, rename this term to `Passage display model attribution` and explicitly scope the fail-closed rule in this document and `docs/passage-display-api.md`.

_Cursor, an AI agent sent on behalf of Austin._

packages/core/src/bible-passage.ts
```diff
@@ -48,6 +48,29 @@ export async function getPassage(
   include_headings?: boolean,
   include_notes?: boolean,
   transform?: boolean,
+): Promise<BiblePassage> {
+  parseBibleVersionId(versionId);
+  await assertUsableVersion(client, versionId);
```
abharms, line 53: issue (non-blocking): restore input validation before filter-related I/O

Moving this await ahead of the existing boolean parsing changes established behavior for malformed JavaScript or unsafe-TypeScript calls. With a language filter, invalid `include_headings` now performs metadata I/O before throwing; for an excluded version, the same call returns `403` instead of the prior validation error.

Suggested change: synchronously parse `versionId`, `include_headings`, and `include_notes` and build the validated request parameters before this await, then pass those values to `getPassageForValidatedVersion`. Do not add new runtime validation for `format`, `transform`, or `usfm` as part of this compatibility fix. Add a regression test asserting an invalid boolean produces the prior Zod error and makes zero client requests when filters are active.

_Cursor, an AI agent sent on behalf of Austin._

packages/core/src/bible-passage-display.ts
cameronapak: Implemented in 3ba975e. Added `scripts/check-cdn-css-major.mjs` to compare the core stylesheet URL with `packages/ui/CDN_CSS_MAJOR_VERSION`, wired it into root `pnpm lint`, and documented the synchronization rule in `PUBLISHING.md`. The guard passes with compatibility major 1.

packages/core/src/bible-passage-display.ts
cameronapak: Implemented in a200ca0. Moved the CSS URL, font ID, and `getBibleStylesheets` into dependency-free `bible-display-resources.ts`, kept the public exports, and added a focused tree-shaking fixture with sentinel and size assertions plus a control import. The helper-only bundle is 0.7 KB and `pnpm check:tree-shaking` passes.

packages/core/src/__tests__/bible-passage-display.test.ts
cameronapak: Implemented in a200ca0 and refined in e02e107. The mocked passage-display suite is skipped whenever `INTEGRATION_TESTS` is set, while the pure `getBibleStylesheets` test still runs. Verification in integration mode reports 1 passed and 10 skipped, with no mocked request able to reach the live API.

packages/core/src/__tests__/bible-passage-display.test.ts
cameronapak: Implemented in a200ca0. Added a repeated same-client, same-options test whose version handler returns different attribution on each request. It asserts both returned values and confirms the version endpoint is called twice.

CONTEXT.md
cameronapak: This contract is scoped to the passage display model rather than every SDK-rendered passage. YPE-4857 specifically targets the core-only partner path, and the ADR introduces a core operation. I updated `CONTEXT.md` and `docs/passage-display-api.md` to make that boundary explicit; `BibleReader` and `BibleCard` retain their existing attribution behavior.

packages/core/src/bible-passage.ts
cameronapak: Implemented in a200ca0. `getPassage` now parses the version ID and builds the validated boolean query before awaiting filter-related I/O, then shares the validated fetch path with the display operation. The regression test passes an invalid boolean with a language filter active and asserts the prior Zod error plus zero requests.

cameronapak:
Ready for re-review

cameronapak:
> ## 🚨 Breaking change detected — signoff required for `v2.14.0`
> This PR adds a changeset declaring a **major** bump, which would ship all three packages as `v2.14.0` (they are a `fixed` group in `.changeset/config.json`, so they version together). Merging is blocked until a repo collaborator with write access comments on this PR with **all four (4) of the following:**
> 
> 1. The verbatim acknowledgment phrase below,
> 2. The precise next version (`v2.14.0` or `2.14.0`),
> 3. The full commit hash being approved (`fb0d209a1a146f3fbd4e54422a3b5895558d1c3f`), and
> 4. A 🚀 (`:rocket:`) emoji.
> 
> I confirm that this is an intentional breaking change, and I have read the release procedures. I understand and have documented its impact upon release.
> 
> **Copy-paste-ready reply:**
> 
> ```
> I confirm that this is an intentional breaking change, and I have read the release procedures. I understand and have documented its impact upon release.
> 
> v2.14.0 fb0d209a1a146f3fbd4e54422a3b5895558d1c3f 🚀
> ```
> 
> The check re-runs automatically when a qualifying comment is posted or edited. Pushing new commits changes the hash, so a fresh signoff is needed.

Wrong. This is not a major bump. 

question: Why is this producing many false positives?

packages/core/src/bible-passage.ts
```diff
@@ -50,28 +72,37 @@ export async function getPassage(
...
+
+async function fetchPassage(
+  client: ApiClient,
+  versionId: number,
+  usfm: string,
+  params: PassageQuery,
+  transform?: boolean,
+): Promise<BiblePassage> {
+  const response = await client.get<unknown>(`/v1/bibles/${versionId}/passages/${usfm}`, params);
+  const passage = BiblePassageSchema.parse(response);
```
Dustin-Kelley: suggestion (non-blocking): `getPassage` now validates responses via `BiblePassageSchema.parse()`

Routing the existing `getPassage()` through `fetchPassage()` means it now runs `BiblePassageSchema.parse()` on every response. Two side effects for a shipped public API:

- `z.object` strips unknown keys, so any field beyond `id`/`content`/`reference` is silently dropped where it was previously passed through.
- A response missing one of those fields now rejects with a `ZodError` instead of returning a partial object.

Fixtures and the documented shape both match the schema, so this is unlikely to break anyone in practice — but it's a behavior change that the changeset (which only describes `getPassageDisplay`) doesn't mention.

Either add a line to `.changeset/fresh-papayas-display.md` noting that `getPassage` now validates responses, or keep the parse on the new display path only and leave `getPassage` returning the raw payload.

---
_Posted by Claude Code on behalf of Dustin Kelley._

cameronapak: Fixed in [5ef2001](https://github.com/youversion/platform-sdk-react/commit/5ef2001432b82548d99bac50affc7ba30e7551f4). `getPassage` once again returns the API payload without schema parsing, preserving unknown fields and its prior partial-response behavior. The new `getPassageDisplay` path still validates with `BiblePassageSchema`, and the regression test covers unknown-field preservation.

Dustin-Kelley (approved):
LGTM!

PR #396, opened 2026-09-11
feat(ui): add Bible Reader search

camrun91:
> [!IMPORTANT]
> This pull request was started by Cameron Llewellyn and passed onto Cameron Pak 

## Bible Reader search

Adds toolbar search with suggestions, browser-local recents, testament filters, verse previews, and navigation that centers the selected verse while dimming surrounding verses. Hosts can use `useBibleSearch` headlessly or override the toolbar action with `onSearchPress`.

Closes YPE-5766. Parent: YPE-5622.

## UI views

Current Chromium captures using live API data. Filters expand as an accordion; reduced motion uses a fade.

| Trending and recents | Suggestions |
| --- | --- |
| Trending and recent searches | Search suggestions before submission |

| Results | Expanded filters |
| --- | --- |
| Verse results with collapsed filters | Old Testament, New Testament, and Both filters |

<details>
<summary>Dark mode</summary>

Dark mode search results and expanded filters

</details>

## Verification and remaining limitations

- Full unit suite: 1,435 passed, 2 skipped. Chromium stories: 61 passed. Build/typecheck and lint passed.
- Inspected live views, keyboard disclosure, reduced motion, and interrupted accordion transitions.
- Some cross-testament searches hit an API pagination HTTP 400; this remains a retryable failure, not a false empty state. A pagination failure can still appear alongside the verse-preview spinner.
- New labels use English fallbacks until upstream localization syncs.

<h2><a href="https://app.greptile.com/api/retrigger?id=63119641"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge.

<h3>Summary</h3>

The PR adds end-to-end Bible Reader search with suggestions, local recents, testament filters, paginated verse results, navigation, and transient verse focus.
- Introduces a reusable `useBibleSearch` session and search client integration.
- Adds the toolbar search interface, result previews, filtering, accessibility behavior, and browser-local recents.
- Centralizes structural USFM parsing with bounded range expansion.
- Adds extensive unit, interaction, navigation, and workflow coverage.

<h3>Diagram</h3>

```mermaid
sequenceDiagram
  participant User
  participant SearchUI as BibleReaderSearch
  participant Hook as useBibleSearch
  participant API as SearchClient
  participant Reader as BibleReader
  User->>SearchUI: Enter or select query
  SearchUI->>Hook: submit query and filter
  Hook->>API: Search verse pages
  API-->>Hook: Hits and continuation token
  Hook-->>SearchUI: Results with previews
  User->>SearchUI: Select verse
  SearchUI->>Reader: Navigate to passage and focus verse
  Reader-->>User: Center verse and dim surroundings
```

<sub>Reviews (16) · Last reviewed commit: ["chore: merge main into Bible search bran..."](https://github.com/youversion/platform-sdk-react/commit/cb1ff011bcb5b2388fe72e81e3cceb00b97f097c)</sub>


packages/ui/src/components/bible-reader.tsx
```diff
@@ -628,6 +633,23 @@ function Root({
   const { books, loading: booksLoading } = useBooks(versionId);
   const booksData = books?.data ?? [];
 
+  const [verseFocus, setVerseFocus] = useState<VerseFocusRequest | null>(null);
+  const verseFocusSeqRef = useRef(0);
+  const navigateToVerse = useCallback(
+    (target: Omit<VerseFocusRequest, 'seq'>) => {
```
cameronapak: **[issue](https://conventionalcomments.org/):** Expose the shared navigation capability so host apps can request a destination before the reader opens.

<details>
<summary>For Agents: public Reader navigation</summary>

> P1, Spec. The [PRD](https://app.notion.com/p/yvproduct/YVP-Search-PRD-for-SDKs-3cff1f2d1b9280d0a346c20c971acf07), “Required public capability” and “Request semantics,” requires a host-to-reader connection that retains one pending request before mount and supports version, chapter/verse/range, passage-only versus full-chapter display, scrolling, and focus.
>
> `navigateToVerse` is created inside `Root` and accepts only book, chapter, and verses. `RootProps` has no navigation connection. Although `useBibleReaderContext` is exported from its source module, it is not re-exported by the package barrel, and a mounted context callback would still not support pre-mount requests or the required options.
>
> Add the public navigation object/ref/hook-equivalent and connect it through `BibleReader.Root`. Retain the newest pending request, consume each request once, and route search selection through the same consumer. Cover pre-mount requests, cross-version navigation, ranges, and no-scroll focus.

</details>

Written by Code Reviewer bot on behalf of Cam.

packages/ui/src/lib/use-transient-verse-focus.ts
```diff
@@ -0,0 +1,118 @@
...
+    const hold = globalThis.setTimeout(clear, SEARCH_VERSE_FOCUS_HOLD_MS);
+    const onInteract = (): void => {
+      clear();
+    };
+    const scroller = overflowAncestor(container);
+    const attachId = globalThis.requestAnimationFrame(() => {
+      container.addEventListener('pointerdown', onInteract);
+      container.addEventListener('keydown', onInteract);
+      scroller.addEventListener('scroll', onInteract);
+    });
```
cameronapak, line 137: **[issue](https://conventionalcomments.org/):** Apply verse focus after automatic scrolling finishes so the scroll does not immediately clear it.

<details>
<summary>For Agents: automatic scroll versus user interaction</summary>

> P1, Spec. The [PRD](https://app.notion.com/p/yvproduct/YVP-Search-PRD-for-SDKs-3cff1f2d1b9280d0a346c20c971acf07), “Scrolling and focus,” requires focus after the scroll lands and clearing on intentional user scrolling.
>
> `scrollIntoView({ behavior: "smooth" })` starts an animation, `paintFocus` runs immediately, and the next animation frame attaches a scroll listener that clears focus. Native smooth scrolling emits those same events. A Chromium reproduction using this exact hook, with no user input, showed focus cleared while the scroller was still near the top; it remained absent when the target landed. The existing jsdom tests stub `scrollIntoView`, so they do not exercise this sequence.
>
> Wait for scroll completion with a bounded fallback before painting focus, and distinguish programmatic scrolling from intentional user interaction. Preserve the reduced-motion path and test a real animated scroll.

</details>

Written by Code Reviewer bot on behalf of Cam.

packages/hooks/src/internal/bible-search-session.ts
```diff
@@ -0,0 +1,366 @@
...
+      verses: parsed.verses,
+    });
+  }
+  return verses;
+}
+
+function flattenHits(pages: readonly VersePage[]): readonly SearchVerseHit[] {
+  const hits: SearchVerseHit[] = [];
+  for (const page of pages) {
+    hits.push(...page.hits);
```
cameronapak, line 309: **[issue](https://conventionalcomments.org/):** Remove repeated verse references when appending results so readers do not see duplicate rows.

<details>
<summary>For Agents: result deduplication</summary>

> P2, Spec. The [PRD](https://app.notion.com/p/yvproduct/YVP-Search-PRD-for-SDKs-3cff1f2d1b9280d0a346c20c971acf07), “Pagination,” requires deduplication by USFM reference across pages and within the incoming page.
>
> `commitPage` deduplicates page tokens only. `flattenHits` concatenates all hits, and `projectVerses` preserves each valid reference. A reducer reproduction with first page [JHN.3.16, ROM.8.28] and second page [ROM.8.28, PSA.23.1, PSA.23.1] returned five rows instead of three. The UI also uses these repeated IDs as sibling React keys.
>
> Deduplicate references while preserving first-seen server order. Cover both an overlapping page boundary and duplicates within one page; token-deduplication tests do not establish this behavior.

</details>

Written by Code Reviewer bot on behalf of Cam.

packages/hooks/src/useBibleSearch.ts
```diff
@@ -0,0 +1,141 @@
...
+  }, [versionId]);
+
+  const { version } = useVersion(versionId);
+  const languageRanges = version?.language_tag ?? null;
+  const debounced = useDebounce(
+    session.lane.kind === 'submitted' ? null : session.normalized,
+    SEARCH_SUGGESTION_DEBOUNCE_MS,
+  );
+  const settled = session.normalized === EMPTY_QUERY ? EMPTY_QUERY : debounced;
+  const qReq = queriesRequest(session, settled);
```
cameronapak, line 69: **[issue](https://conventionalcomments.org/):** Ignore an earlier suggestion response once the user has changed the query.

<details>
<summary>For Agents: suggestion identity during debounce</summary>

> P2, Spec. The [PRD](https://app.notion.com/p/yvproduct/YVP-Search-PRD-for-SDKs-3cff1f2d1b9280d0a346c20c971acf07), “Debounce and request coordination,” requires checking the current normalized query before applying a response.
>
> `qReq` and the active query key follow the debounced value rather than the current normalized input. In a controlled hook reproduction, I started the suggestion request for "old", typed "new", then resolved the old request during the new 300 ms debounce. The hook returned query "new" with the newly arrived "old result" suggestion. This is a late response changing the UI, not merely retaining a previously completed list.
>
> Supersede the old active request as soon as normalized input changes, or explicitly gate its response against the current input identity. Add a deferred-response test that resolves the earlier request during the next debounce window.

</details>

Written by Code Reviewer bot on behalf of Cam.

packages/ui/src/components/bible-reader-search.tsx
```diff
@@ -0,0 +1,388 @@
...
+  versionAbbreviation: string;
+  onSelect: () => void;
+}): ReactElement {
+  const { passage, loading, error } = usePassage({
+    versionId,
+    usfm: verse.id,
+    format: 'text',
+    include_headings: false,
+    include_notes: false,
+    options: { keepPreviousData: false },
```
cameronapak, line 623: **[issue](https://conventionalcomments.org/):** Wait until result rows approach visibility before loading their verse text.

<details>
<summary>For Agents: visibility-gated passage enrichment</summary>

> P2, Spec. The [PRD](https://app.notion.com/p/yvproduct/YVP-Search-PRD-for-SDKs-3cff1f2d1b9280d0a346c20c971acf07), “Result rendering,” says to load text lazily as rows become visible.
>
> `VerseResults` mounts every result through `verses.map`, and every `SearchVerseRow` enables `usePassage` immediately. An uncached initial page therefore starts passage requests for all 20 rows, including rows outside the scroll viewport. Per-reference caching does not prevent first-time off-screen requests.
>
> Use row visibility to enable enrichment, with a modest prefetch margin if needed. The existing `usePassage` API already accepts `options.enabled`; keep the reference row selectable before text arrives and after an enrichment failure. Verify that off-screen rows do not fetch until they approach visibility.

</details>

Written by Code Reviewer bot on behalf of Cam.

packages/ui/src/components/bible-reader-search.tsx
```diff
@@ -0,0 +1,388 @@
...
+              versionId={versionId}
+              bookName={booksData.find((book) => book.id === verse.book)?.title ?? verse.book}
+              versionAbbreviation={versionAbbreviation}
+              onSelect={() => onSelect(verse)}
+            />
+          </li>
+        ))}
+      </ul>
+      {nextPage === 'available' ? (
+        <Button variant="secondary" onClick={onLoadMore}>
```
cameronapak: **[issue](https://conventionalcomments.org/):** Load the next page as the reader approaches the end of the results.

<details>
<summary>For Agents: automatic continuation loading</summary>

> P2, Spec. The [PRD](https://app.notion.com/p/yvproduct/YVP-Search-PRD-for-SDKs-3cff1f2d1b9280d0a346c20c971acf07), “Pagination,” requires requesting the next page when the user approaches the last five rows.
>
> The only `onLoadMore` invocation here is the button click. No row-visibility or scroll handler requests a continuation page, so scrolling to the end never loads additional results.
>
> Trigger the existing guarded `loadMore` action when a near-end row becomes visible and `nextPage` is available. Keep the explicit retry action for failures. Verify automatic loading, one request at a time, and stopping after token exhaustion.

</details>

Written by Code Reviewer bot on behalf of Cam.

packages/hooks/src/useBibleSearch.ts
```diff
@@ -0,0 +1,141 @@
...
+  const client = useSearchClient();
+  const keyBase = useQueryKeyBase();
+  const [session, dispatch] = useReducer(searchSessionReducer, versionId, createSearchSession);
+
+  useEffect(() => {
+    dispatch({ type: 'setVersion', versionId });
+  }, [versionId]);
+
+  const { version } = useVersion(versionId);
+  const languageRanges = version?.language_tag ?? null;
```
cameronapak: **[issue](https://conventionalcomments.org/):** Use the all-languages fallback when the selected Bible has no usable language information.

<details>
<summary>For Agents: wildcard language fallback</summary>

> P2, Spec. The [PRD](https://app.notion.com/p/yvproduct/YVP-Search-PRD-for-SDKs-3cff1f2d1b9280d0a346c20c971acf07), “Search-query endpoint” and its acceptance scenarios, requires the wildcard language range when no usable tag is available.
>
> When `useVersion` fails, `languageRanges` remains null and `enabled` keeps discovery disabled indefinitely. A hook reproduction with rejected version metadata recorded zero trending calls rather than a call with "*". An empty `language_tag` is also passed through rather than replaced with a usable range.
>
> Distinguish metadata still loading from metadata unavailable. Once loading settles, use a valid language tag or "*" for trending and suggestions. Cover failed metadata and empty/unusable tags without converting suggestion failures into the verse-search error state.

</details>

Written by Code Reviewer bot on behalf of Cam.

packages/ui/src/components/bible-reader-search.tsx
```diff
@@ -0,0 +1,388 @@
...
+            theme={background}
+            className="yv:flex yv:max-h-[min(36rem,80vh)] yv:w-[calc(100vw-2rem)] yv:max-w-md yv:flex-col yv:gap-4 yv:p-4"
+          >
+            <DialogTitle className="yv:text-base yv:font-semibold">
+              {t('bibleSearchDialogTitle', 'Search')}
+            </DialogTitle>
+            <DialogDescription className="yv:sr-only">
+              {t('bibleSearchPlaceholder', 'Search verses')}
+            </DialogDescription>
+            <SearchPanel onClose={() => setOpen(false)} />
```
cameronapak: **[issue](https://conventionalcomments.org/):** Add a visible Close or Done action so people can dismiss search without selecting a result.

<details>
<summary>For Agents: explicit dialog dismissal</summary>

> P2, Spec. The [PRD](https://app.notion.com/p/yvproduct/YVP-Search-PRD-for-SDKs-3cff1f2d1b9280d0a346c20c971acf07), “Entry and presentation,” explicitly requires a localized Done/Close action.
>
> This dialog contains only the title, description, and search panel. Shared `DialogContent` does not insert a close control. The rendered browser accessibility tree confirmed there was no Close/Done action; dismissal currently relies on Escape, an outside click, or result selection.
>
> Add a localized close button that invokes the dialog open-state handler. Verify keyboard and pointer dismissal, including when the query and results are empty.

</details>

Written by Code Reviewer bot on behalf of Cam.

packages/ui/src/components/bible-reader-search.tsx
```diff
@@ -0,0 +1,388 @@
...
+    case 'empty':
+      return (
+        <p className="yv:text-sm yv:text-muted-foreground">
+          {t('bibleSearchNoVerseResults', 'No verses matched this search.')}
+        </p>
+      );
+    case 'failed':
+      return (
+        <div className="yv:flex yv:flex-col yv:gap-2">
+          <p className="yv:text-sm yv:text-muted-foreground">
```
cameronapak, line 424: **[issue](https://conventionalcomments.org/):** Announce search failures so screen-reader users know the search failed and can retry.

<details>
<summary>For Agents: search error alert semantics</summary>

> P2, Spec. The [PRD](https://app.notion.com/p/yvproduct/YVP-Search-PRD-for-SDKs-3cff1f2d1b9280d0a346c20c971acf07), “Accessibility, localization, and internationalization,” requires alert semantics for error states.
>
> The asynchronous failure branch inserts a plain paragraph, with neither `role="alert"` nor a live region. Focus remains elsewhere in the dialog, so the new failure message is not exposed as an announcement. `LoaderIcon` already supplies status semantics for loading; that does not announce the later error.
>
> Apply the established alert pattern used by `VerseUnavailableMessage` in `verse.tsx`, and assert the accessible error state when a submitted search fails. Keep the retry button available.

</details>

Written by Code Reviewer bot on behalf of Cam.

packages/hooks/src/internal/bible-search-session.ts
```diff
@@ -0,0 +1,366 @@
...
+export const MAX_SEARCH_QUERY_LENGTH = 100;
+export const SEARCH_SUGGESTION_DEBOUNCE_MS = 300;
+export const SEARCH_VERSES_PAGE_SIZE = 20;
+export const EMPTY_QUERY =
+  // SAFETY: the empty string is the branded empty query. normalizeQuery is the
+  // only other mint, and it returns this constant after trim.
+  '' as NormalizedQuery;
+
+export function clampSearchInput(raw: string): string {
+  return raw.length <= MAX_SEARCH_QUERY_LENGTH ? raw : raw.slice(0, MAX_SEARCH_QUERY_LENGTH);
```
cameronapak: **[issue](https://conventionalcomments.org/):** Count visible characters when limiting the query so composed text is not cut short.

<details>
<summary>For Agents: user-perceived character limit</summary>

> P2, Spec. The [PRD](https://app.notion.com/p/yvproduct/YVP-Search-PRD-for-SDKs-3cff1f2d1b9280d0a346c20c971acf07), “Search input,” specifies a maximum of 100 user-perceived characters.
>
> `raw.length` and `slice(0, 100)` count UTF-16 code units. The input also has native `maxLength={100}`, which uses that same unit. A reproduction using 60 composed accented characters ("a\u0301" repeated) was truncated to 50 visible characters. Headless input can also be cut inside a character cluster.
>
> Clamp by grapheme cluster, for example with `Intl.Segmenter`, and remove the stricter code-unit limit from the DOM input. Reconcile the existing Core query validator with the same contract so the UI does not accept a query that Core rejects. Cover combining marks and surrogate-pair characters around the boundary.

</details>

Written by Code Reviewer bot on behalf of Cam.

packages/core/src/usfm-reference.ts
```diff
@@ -0,0 +1,72 @@
...
+    return { book, chapter, verses: [verseStart] };
+  }
+
+  const verseEnd = Number(match[4]);
+  if (!isPositiveSafeInteger(verseEnd) || verseEnd < verseStart) {
+    return null;
+  }
+
+  if (verseEnd - verseStart + 1 > MAX_RANGE_LENGTH) {
+    return null;
```
cameronapak, line 64: **[praise](https://conventionalcomments.org/):** Bounding verse ranges before expanding them prevents malformed results from hanging the reader.

<details>
<summary>For Agents: safe reference parsing</summary>

> `Number.isSafeInteger`, the reversed-range check, and the maximum span bound reject unsafe inputs before the expansion loop. Schema validation and result projection share this parser, and the tests cover unsafe endpoints and oversized ranges. The targeted parser and verse-search run passed all 34 tests.

</details>

Written by Code Reviewer bot on behalf of Cam.

cameronapak (dismissed):
## Review

[YPE-5766](https://lifechurch.atlassian.net/browse/YPE-5766)

## Summary

Standards: 0 must-fix. Spec: 10 must-fix. Primary concern: host navigation is unavailable and automatic scrolling clears selected-verse focus before arrival.

<details>
<summary>For Agents</summary>

> - CI: failing. Integration Tests reports 582 passing assertions and 11 unhandled rejections; the cause is not established. Other build, typecheck, lint, and unit-test checks pass.
> - Bot review: clear; earlier bot findings are resolved.
> - Event: REQUEST_CHANGES
> - HEAD: 9720ea4332789a51fd6e01e6b08d1d6062a8f50f
> - Spec: the user-supplied artifact of the [Search PRD](https://app.notion.com/p/yvproduct/YVP-Search-PRD-for-SDKs-3cff1f2d1b9280d0a346c20c971acf07), linked from parent [YPE-5622](https://lifechurch.atlassian.net/browse/YPE-5622).
> - Verification: 92 existing targeted Vitest tests passed. Three focused regression reproductions failed for result deduplication, stale suggestions, and wildcard fallback. Chromium confirmed automatic scrolling clears focus before arrival. Temporary reproduction fixtures were removed.

</details>

Written by Code Reviewer bot on behalf of Cam.

packages/ui/src/components/bible-reader.tsx
cameronapak: Fixed in [39fffb5](https://github.com/youversion/platform-sdk-react/commit/39fffb5). `BibleReaderNavigation` is now public and connects through `BibleReader.Root`. It retains the newest pre-mount request, consumes it once, and shares its consumer with search selection. Tests cover StrictMode, remounts, cross-version ranges, passage/full-chapter display, and in-place no-scroll focus.

packages/ui/src/lib/use-transient-verse-focus.ts
cameronapak: Fixed in [39fffb5](https://github.com/youversion/platform-sdk-react/commit/39fffb5). Focus now waits for scroll completion, with a bounded fallback, before starting its hold. User input cancels focus; programmatic scroll events do not. The Chromium story verifies a real smooth scroll to John 1:51, focus retained after arrival, and wheel cancellation. Reduced-motion coverage also passes.

packages/hooks/src/internal/bible-search-session.ts
cameronapak: Fixed in [39fffb5](https://github.com/youversion/platform-sdk-react/commit/39fffb5). `projectVerses` now deduplicates passage IDs in first-seen server order. Regression tests cover duplicates within a page and across an overlapping page boundary.

packages/hooks/src/useBibleSearch.ts
cameronapak: Fixed in [39fffb5](https://github.com/youversion/platform-sdk-react/commit/39fffb5). Suggestion requests are disabled as soon as the current normalized input differs from the debounced identity. A deferred-response regression test resolves the earlier request during the next debounce window and verifies that its suggestions do not appear.

packages/ui/src/components/bible-reader-search.tsx
cameronapak: Fixed in [39fffb5](https://github.com/youversion/platform-sdk-react/commit/39fffb5). Row visibility now gates `usePassage` through `options.enabled`, with a 200px prefetch margin. Reference buttons remain selectable without enrichment. The component test verifies enrichment is disabled before intersection and enabled afterward.

packages/ui/src/components/bible-reader-search.tsx
cameronapak: Fixed in [39fffb5](https://github.com/youversion/platform-sdk-react/commit/39fffb5). Intersection with any of the final five rows now triggers guarded continuation loading. Tests cover jumping to the bottom, duplicate observer callbacks, an active request, duplicate-only continuation pages, and token exhaustion. Explicit retry remains available after failure.

packages/hooks/src/useBibleSearch.ts
cameronapak: Fixed in [39fffb5](https://github.com/youversion/platform-sdk-react/commit/39fffb5). Discovery waits while version metadata loads, then uses a validated language range or `*` when metadata is unavailable or unusable. Hook tests pass for failed metadata and empty or invalid language tags.

packages/ui/src/components/bible-reader-search.tsx
cameronapak: Fixed in [39fffb5](https://github.com/youversion/platform-sdk-react/commit/39fffb5). The dialog now has a visible Close button using the existing localized `closeAriaLabel` key and its open-state handler. Keyboard dismissal is covered by a component test; pointer dismissal and the rendered control were also checked.

packages/ui/src/components/bible-reader-search.tsx
cameronapak: Fixed in [39fffb5](https://github.com/youversion/platform-sdk-react/commit/39fffb5). Submitted-search and continuation failures now expose `role="alert"` and retain their retry controls. Component coverage and the Chromium SearchFailure story pass, including the accessible alert and enabled retry button.

packages/hooks/src/internal/bible-search-session.ts
cameronapak: Fixed in [39fffb5](https://github.com/youversion/platform-sdk-react/commit/39fffb5). Core now owns the shared 100-grapheme clamp using `Intl.Segmenter`; the Core validator and hooks use the same contract. The input no longer applies a UTF-16 `maxLength`. Boundary tests pass for combining marks and surrogate-pair characters.

.size-limit.js
cameronapak, line 1: This change allows for us to be able to easily add Zod to the UI package

Dustin-Kelley (commented):
Review posted by Claude (AI agent) on behalf of Dustin. Four inline comments below using Conventional Comments.

packages/core/src/schemas/search.ts
```diff
@@ -1,16 +1,29 @@
 import * as z from 'zod/mini';
+import { parseUsfmReference } from '../usfm-reference';
+
+export const MAX_SEARCH_QUERY_GRAPHEMES = 100;
+
+const searchTextSegmenter = new Intl.Segmenter(undefined, { granularity: 'grapheme' });
```
Dustin-Kelley: **issue (blocking):** `new Intl.Segmenter(...)` runs at module-evaluation time in the core barrel.

Any runtime without `Intl.Segmenter` (React Native/Hermes, older Firefox ESR) throws `TypeError: Intl.Segmenter is not a constructor` on `import '@youversion/platform-core'` — which also takes down `platform-react-hooks` since it re-exports core. That breaks auth, Bible clients, everything, not just search.

Suggest constructing the segmenter lazily inside `clampSearchText` and falling back to `Array.from(raw)` (code points) when `Intl.Segmenter` is undefined, so the cost stays confined to search.

_Posted by Claude on behalf of Dustin._

packages/hooks/src/utility/useDebounce.ts
```diff
@@ -16,12 +16,17 @@ export function useDebounce<T>(value: T, delay: number): T {
   const [debouncedValue, setDebouncedValue] = useState(value);
 
   useEffect(() => {
+    if (value === null) {
```
Dustin-Kelley: **issue (non-blocking):** `useDebounce` is a public export from the hooks barrel, and the new `value === null` short-circuit silently changes its contract for every existing consumer.

A caller doing `useDebounce<string | null>(hoveredVerseId, 200)` to avoid tooltip flicker now sees `null` land synchronously after this minor bump. Neither the docstring nor the changeset mentions it.

Suggest keeping the generic hook as-is and doing the submitted-lane short-circuit in `useBibleSearch` — debounce the string and ignore the result while `lane.kind === 'submitted'`.

_Posted by Claude on behalf of Dustin._

packages/hooks/src/internal/bible-search-session.ts
```diff
@@ -0,0 +1,393 @@
...
+  raw: string,
+  normalized: NormalizedQuery,
+): SearchSession {
+  if (normalized === EMPTY_QUERY) {
+    if (state.raw === '' && state.normalized === EMPTY_QUERY && state.lane.kind === 'browsing') {
+      return state;
+    }
+    return { ...state, raw: '', normalized: EMPTY_QUERY, lane: { kind: 'browsing' } };
+  }
+  if (state.lane.kind === 'submitted' && state.lane.query === normalized) {
```
Dustin-Kelley: **issue (non-blocking):** re-submitting the same normalized query is a reducer no-op, so Enter after a failed or empty search does nothing.

Scenario: search "love" → request fails → phase `failed`. User presses Enter again expecting a retry → `submitNormalized` returns the unchanged state, no refetch. Only the separate Try again button works. Same branch also means `raw` isn't refreshed when a picked suggestion differs from the current query only by whitespace.

Consider resetting `pages`/`wantsMore` (or routing to retry) when the same query is resubmitted.

_Posted by Claude on behalf of Dustin._

packages/hooks/src/useBibleSearch.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  const vReq = versesRequest(session);
+  const verses = useApiData(
+    [
+      ...keyBase,
+      'searchVerses',
+      vReq?.versionId,
+      vReq?.query ?? '',
+      vReq?.pageToken ?? '',
+      props.bookIds,
```
Dustin-Kelley, line 95: **suggestion (non-blocking):** `props.bookIds` is part of the `searchVerses` query key, but the fetch never uses it — filtering is client-side in `deriveSearchPhase`.

Toggling the testament filter while a page is in flight changes the key, so TanStack abandons that request and fires an identical `/v1/search-verses` call under the new key; the first response is cached under a key that's never observed again.

Dropping `props.bookIds` from the key removes the redundant request with no behaviour change.

_Posted by Claude on behalf of Dustin._

packages/hooks/src/utility/useDebounce.ts
abharms: Duplicate of the earlier `useDebounce` thread: https://github.com/youversion/platform-sdk-react/pull/396#discussion_r4048339965. Superseded and resolved.

abharms (commented):
Reviewed the current head (`77aa9e6`) against YPE-5766, its parent ticket, the PR's stated behavior, the public navigation contract, and the existing review threads. I left four confirmed issues inline; the first three are blocking.

The exclusive browsing/submitted search lane and stale-response guards are especially well designed—they make concurrent suggestion/result behavior much easier to reason about. The transient-focus implementation also shows thoughtful accessibility work around focus timing, reduced motion, and preserving user input.

I did not rerun the suite locally; the PR's GitHub checks are green at this head.

packages/ui/src/components/bible-reader.tsx
```diff
@@ -635,13 +647,49 @@ function Root({
   const { books, loading: booksLoading } = useBooks(versionId);
   const booksData = books?.data ?? [];
 
+  const [verseFocus, setVerseFocus] = useState<VerseFocusRequest | null>(null);
+  const verseFocusSeqRef = useRef(0);
+  const [localNavigation] = useState(() => new BibleReaderNavigation());
+  const navigation = navigationProp ?? localNavigation;
+  useEffect(() => {
```
abharms, line 688: issue: A request queued before mount is documented as being consumed on mount, but consuming it only in a passive effect means the reader children first render with the default/controlled location. `Content` can therefore start a request for that passage (and potentially show cached content) before this effect switches to the queued destination. Could the pending destination be exposed non-destructively so it can seed the initial reader state before children render, then be consumed after commit? A test asserting that the default passage is never requested for a pre-mount navigation would catch this.

packages/ui/src/components/bible-reader.tsx
```diff
@@ -717,7 +769,17 @@ function Content() {
     return booksData.find((b) => b.id === book);
   }, [booksData, book]);
 
-  const usfmReference = `${book}.${chapter}`;
+  const chapterReference = `${book}.${chapter}`;
+  const activeNavigation =
```
abharms, line 810: issue (blocking): Controlled prop navigation can reactivate an old verse focus. After focusing `JHN.3.16`, if the host changes controlled `chapter` to `4` and later back to `3`, `verseFocus` was never cleared, so this condition becomes true again and `usfmReference` reverts to `JHN.3.16` instead of the full chapter. The context setter wrappers clear focus, but direct controlled-prop changes bypass them. Could we clear an activated focus when the externally controlled destination leaves it, while preserving a genuinely pending cross-destination request?

packages/hooks/src/internal/bible-search-session.ts
```diff
@@ -0,0 +1,393 @@
...
+  for (const hit of hits) {
+    if (seen.has(hit.id)) {
+      continue;
+    }
+    seen.add(hit.id);
+    const parsed = parseUsfmReference(hit.id);
+    if (parsed === null) {
+      continue;
+    }
+    verses.push({
```
abharms, line 296: issue (blocking): Chapter-only API hits remain selectable results here (`verses: []`), but every result selection calls `focusReference`. The transient-focus path exits when there are no verses, so selecting one closes search and opens the chapter without the selected-verse centering/focus/dimming promised by this feature. Could we either filter chapter-only hits out of Bible results (matching the reference implementation's verse-only display) or render them as explicitly different chapter-navigation rows with their own behavior and test?

packages/core/src/schemas/search.ts
cameronapak: > Fixed in [7d6fe84](https://github.com/youversion/platform-sdk-react/commit/7d6fe8440aeed8704d7a58a3dda24680b3fcce86). Core no longer constructs Intl.Segmenter on import. Search constructs it lazily and falls back to Unicode code points when unavailable.
>
> <details>
> <summary>Verification evidence</summary>
>
> - Coordinator-run regression in `packages/core/src/__tests__/search-verses.test.ts` removes Intl.Segmenter before importing the core barrel, then checks truncation and schema validation at the surrogate-pair boundary. Existing native grapheme tests remain passing.
> - At the linked revision, `pnpm test --env-mode=loose` with test environment configuration passed 1,437 tests; 2 live API tests were skipped. Typecheck and lint passed after correcting the test's schema import and capability-check syntax.
> - Limit: this simulates the missing feature in Node, not a Hermes device. The fallback counts code points, not complete grapheme clusters.
>
> </details>
>
> Written by Coding Agent on behalf of Cam.

packages/hooks/src/utility/useDebounce.ts
cameronapak: > Fixed in [7d6fe84](https://github.com/youversion/platform-sdk-react/commit/7d6fe8440aeed8704d7a58a3dda24680b3fcce86). The generic useDebounce hook delays null again. Search debounces its normalized string and keeps suggestion suppression in the search-session guards.
>
> <details>
> <summary>Verification evidence</summary>
>
> - Coordinator-run coverage in `packages/hooks/src/utility/useDebounce.test.tsx` checks null before and at the delay boundary, plus cancellation when another value supersedes it.
> - `packages/hooks/src/useBibleSearch.test.tsx` verifies that submitted searches do not request suggestions and that edits do not display stale suggestions during debounce.
> - At the linked revision, `pnpm test --env-mode=loose` with test environment configuration passed 1,437 tests, including these regressions; 2 live API tests were skipped. Typecheck and lint passed after correcting the timer-test callbacks.
>
> </details>
>
> Written by Coding Agent on behalf of Cam.

packages/hooks/src/internal/bible-search-session.ts
cameronapak: > Fixed in [7d6fe84](https://github.com/youversion/platform-sdk-react/commit/7d6fe8440aeed8704d7a58a3dda24680b3fcce86). Enter can resubmit failed or empty searches without replaying cached pages. Repeated submissions while loading share the active request, and whitespace-only suggestion changes update the input.
>
> <details>
> <summary>Verification evidence</summary>
>
> - Coordinator-run hook coverage exercises failed → empty → pending → results, checks request counts during repeated Enter, and verifies the updated raw input. Reducer coverage verifies that resubmission clears pages.
> - At the linked revision, `pnpm test --env-mode=loose` with test environment configuration passed 1,437 tests; 2 live API tests were skipped. `STORYBOOK_YOUVERSION_APP_KEY=mock-key STORYBOOK_YOUVERSION_API_HOST=api.youversion.com STORYBOOK_AUTH_REDIRECT_URL=http://localhost:6006 pnpm --filter @youversion/platform-react-ui exec vitest run --project storybook` passed 61 Chromium stories.
> - Live-browser check: blocked the search request for love, observed the error, unblocked it, then pressed Enter. John 3:16 loaded and the alert disappeared.
> - Limit: empty resubmission is verified by the hook test. The browser empty-state fixture did not reach its expected state and is not counted as verification.
>
> </details>
>
> Written by Coding Agent on behalf of Cam.

packages/hooks/src/useBibleSearch.ts
cameronapak: > I'm keeping bookIds in this query key for this PR. You're right that it can duplicate a request, but removing it changes the agreed filter behavior: switching testament after a failed continuation must immediately clear the stale error and retry. Avoiding the extra request needs separate error-reset/retry handling rather than only removing the key.
>
> <details>
> <summary>Verification evidence</summary>
>
> - Coordinator-run counterexample: temporarily removed bookIds and ran pnpm --filter @youversion/platform-react-hooks exec vitest run src/useBibleSearch.test.tsx -t 'replaces a failed continuation'. The test failed: expected nextPage: loading, received nextPage: failed.
> - Restored bookIds in https://github.com/youversion/platform-sdk-react/commit/7d6fe8440aeed8704d7a58a3dda24680b3fcce86. Full unit suite: 1437 passed, 2 live API tests skipped; 61 Chromium stories passed.
> - This retains the extra-request tradeoff; it does not implement request deduplication across testament filters.
>
> </details>
>
> Written by Coding Agent on behalf of Cam.

packages/ui/src/components/bible-reader.tsx
cameronapak: > Fixed in [c6e5efc](https://github.com/youversion/platform-sdk-react/commit/c6e5efc406b871248354e5fface77ccc7ab0c9a3). Queued navigation is read without consuming during render, so the first passage request uses the queued destination. Controlled readers wait for the host to accept that destination, and consumption still happens once after commit.
>
> <details>
> <summary>Verification evidence</summary>
>
> - Coordinator-run: `YVP_APP_KEY=mock-test-key YVP_API_HOST=api.youversion.com pnpm --filter @youversion/platform-react-ui exec vitest run --project unit src/components/bible-reader-navigation.test.tsx` passed all 3 tests. The StrictMode test checks every initial passage request, not just the final request; the controlled case checks that no passage request is enabled before staggered host updates reach the target. Newest-request, consume-once, remount, range, cross-version, and in-place focus coverage also passes.
> - At the linked revision, the full unit suite passed 1,438 tests with 2 live API skips. An initial typecheck failure in the new test cleanup callback was corrected to return void; typecheck and the final unit rerun passed.
> - Chromium inspection confirmed queued John 1:51 navigation focused that verse and dimmed surrounding verses. Verification used mocks, not live API, Hermes, or Safari; it does not establish the state of new CI.
>
> </details>
>
> Written by Coding Agent on behalf of Cam.

packages/ui/src/components/bible-reader.tsx
cameronapak: > Fixed in [c6e5efc](https://github.com/youversion/platform-sdk-react/commit/c6e5efc406b871248354e5fface77ccc7ab0c9a3). Leaving an activated destination through controlled props now clears its navigation state. Returning opens the full chapter without restoring the old passage or focus; requests still waiting for the host remain pending.
>
> <details>
> <summary>Verification evidence</summary>
>
> - Coordinator-run: `YVP_APP_KEY=mock-test-key YVP_API_HOST=api.youversion.com pnpm --filter @youversion/platform-react-ui exec vitest run --project unit src/components/bible-reader-navigation.test.tsx` passed all 3 tests. The controlled workflow requests passage-only `JHN.3.16`, leaves for chapter 4, then returns and verifies the full `JHN.3` request. It also covers a pending cross-version focus request, staggered host acceptance, and clearing focus state after leaving and returning to that version.
> - At the linked revision, the full unit suite passed 1,438 tests with 2 live API skips. An initial typecheck failure in the new test cleanup callback was corrected to return void; typecheck and the final unit rerun passed.
> - This controlled-prop regression is verified with component tests and hook overrides, not a live host application. No live API, Hermes, Safari, or new-CI result is claimed.
>
> </details>
>
> Written by Coding Agent on behalf of Cam.

packages/hooks/src/internal/bible-search-session.ts
cameronapak: > Fixed in [c6e5efc](https://github.com/youversion/platform-sdk-react/commit/c6e5efc406b871248354e5fface77ccc7ab0c9a3). Chapter-only API hits are omitted from Bible search results, matching Swift. Verse ranges remain supported, and pages with no selectable verses continue until pagination is exhausted.
>
> <details>
> <summary>Verification evidence</summary>
>
> - Coordinator-run: `YVP_APP_KEY=mock-test-key YVP_API_HOST=api.youversion.com pnpm --filter @youversion/platform-react-hooks exec vitest run src/internal/bible-search-session.test.ts src/useBibleSearch.test.tsx --maxWorkers=2` passed 44 tests. Projection coverage retains `JHN.3.16-18` while dropping `JHN.3`; phase coverage keeps searching through nonmatching pages and reports empty only after exhaustion.
> - With mock Storybook configuration, `pnpm --filter @youversion/platform-react-ui exec vitest run --project storybook --maxWorkers=2` passed 61 Chromium stories. `SearchAndReturn` now receives mixed chapter/verse hits and checks that only the verse is selectable; `EmptyResults` receives a chapter-only hit and checks the empty state rather than an error or spinner.
> - At the linked revision, browser DOM inspection also confirmed the chapter-only empty state, zero chapter rows, and zero alerts; its screenshot was inspected. These checks use mocked HTTP, not live API, Hermes, or Safari, and do not establish the state of new CI.
>
> </details>
>
> Written by Coding Agent on behalf of Cam.

abharms (approved):
🫡

PR #417, opened 2026-09-17
fix(ui): honor auth button mode

cameronapak:
## Summary
Fix a mismatch in `YouVersionAuthButton`: an authenticated user could see **Sign in**, but clicking the button signed them out.

```diff
 mode: `signIn` or omitted, user: authenticated
 
 label: Sign in
-click: Sign out
+click: Sign in
```

The button now uses one mode-aware decision for both its label and click action:

```text
signIn or omitted  -> sign in
signOut            -> sign out
auto               -> follow the user's authentication state
```

This keeps explicit and default sign-in buttons from silently signing out authenticated users. Existing `auto` behavior does not change.

## Who this affects
Consumers that render `YouVersionAuthButton` while a user session already exists.

## Verification
- Adds a component test covering explicit `signIn`, default mode, and authenticated `auto` mode
- GitHub build, test, integration, lint, and type-check checks pass


camrun91 (approved):
LGTM

